PR #4, opened 2026-04-21
fix: remove redundant DeployExtension target from SignalR csproj

larry-fox-lobster:
## Problem
The SignalR extension csproj contained a `DeployExtension` MSBuild target that copied built files to `~/.botnexus/extensions/signalr/`. This duplicated work already handled by the CLI and used $(USERPROFILE), a Windows-only variable that breaks on Linux.

## Fix
Removed the entire `DeployExtension` target and its associated comment. The CLI handles extension deployment.


sytone:
@copilot Looking at the security scan it is flagging workflows and skills related to security checking and validation. So these are false positives. Can these file be excluded from the security scan that is running as part of the pr check

PR #70, opened 2026-04-29
fix(portal): expand only targeted config section when navigating via URL fragment

### LINKED ISSUE #69: [Portal] Config page shows all sections when navigating to a specific one
Clicking a config section link (e.g. Cron from the sidebar) scrolls to that section but all other sections are still visible. Expected: navigating to #cfg-cron should expand only the Cron section and collapse all others.

Fix: Configuration.razor should read the URI fragment on load and navigation, expand only the matching section, and collapse the rest. ConfigSection needs a way to receive its expanded state from the parent rather than managing it independently.

larry-fox-lobster:
Closes #69

Navigating to `configuration#cfg-cron` now expands only Cron and collapses everything else. No fragment = Gateway expanded by default. User can still toggle sections manually.


larry-fox-lobster:
Superseded by #71. The section expansion/collapse mechanism is being replaced by dedicated panels loaded per navigation click.

PR #118, opened 2026-05-03
fix(telegram): self-load config from config.json when IOptions is empty at startup

larry-fox-lobster:
## Problem

Extension adapters (including Telegram) are loaded via a custom `AssemblyLoadContext` **after** DI service registration completes. This means `IOptions<TelegramGatewayOptions>` is always resolved to an empty default instance — the `channels.telegram` config section is never bound.

Result: `Telegram bot 'default' requires BotToken` on every gateway start.

## Fix

`TelegramChannelAdapter.EnsureBotsInitialized()` now falls back to reading `channels.telegram` directly from `~/.botnexus/config.json` when `IOptions` is empty (no `BotToken` and no `Bots`).

Also adds `GatewayServiceCollectionExtensions.BindChannelOptions()` which attempts post-registration binding for forward-compatibility.

## Tests
3 new tests in `ChannelOptionsBindingTests`

### DIFF
--- a/src/extensions/BotNexus.Extensions.Channels.Telegram/TelegramChannelAdapter.cs
+++ b/src/extensions/BotNexus.Extensions.Channels.Telegram/TelegramChannelAdapter.cs
@@ -2,6 +2,7 @@ using System.Collections.Concurrent;
 using System.Globalization;
 using System.Net.Http;
 using System.Text;
+using System.Text.Json;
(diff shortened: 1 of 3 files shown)


larry-fox-lobster:
Superseded — migrating to IConfiguration instead. This eliminates the root cause rather than working around it.

PR #149, opened 2026-05-05
fix(cli): update command hangs — add 60s timeout, drain stdout/stderr, suppress credential prompts

larry-fox-lobster:
Fixes a hang in `botnexus update` during the `git pull` step.

## Root causes

1. **No timeout** — if the network is slow, git is waiting for SSH keys, or there's a credential prompt, `git pull` blocks indefinitely with no way to escape short of Ctrl+C.

2. **Pipe buffer deadlock** — `RedirectStandardOutput = !verbose` redirected the pipe but never read it. A process writing enough output to fill the OS pipe buffer (~64KB) would block forever waiting for the reader that never comes.

## Fix

- Always redirect and drain stdout/stderr. Verbose mode streams to console; silent mode discards.
- 60-second hard timeout via linked `CancellationTokenSource`. Kills the process tree on expiry with a clear error message.
- `GIT_TERMINAL_PROMPT=0` + `SSH BatchMode=yes` — fail fast on auth issues instead of hanging on a credential prompt.
- `--no-rebase` to avoid interactive merge/rebase prompts.

### DIFF
--- a/src/gateway/BotNexus.Cli/Commands/BuildOutputStreamer.cs
+++ b/src/gateway/BotNexus.Cli/Commands/BuildOutputStreamer.cs
@@ -31,6 +31,11 @@ internal static partial class BuildOutputStreamer
             RedirectStandardError = true,
             CreateNoWindow = true
         };
+        // Disable telemetry and skip first-run messages that can add latency.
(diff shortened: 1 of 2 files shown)


larry-fox-lobster:
Superseded by #151 which applies the same fix on top of the visual polish (#150).

PR #201, opened 2026-05-11
fix: keep sub-agent output in originating conversation

sytone:
## Summary

Fixes internal wake messages creating and routing through `internal:` conversations, which caused sub-agent output to appear in hidden/wrong conversations instead of the one the user was looking at.

## Changes

- **Routes sub-agent lifecycle and final output back to the originating (default) conversation** instead of letting internal wakeup messages spawn separate `internal:` threads.
- **Gateway keeps internal wakeups on the parent session**, preventing them from leaking into user-visible conversation state.
- **Sub-agent run views remain visible as read-only inspection views** while `internal:` runtime threads are hidden from the normal conversation list.
- **Sub-agent workspaces are moved to temporary isolated locations** and cleaned up automatically on completion, failure, timeout, or kill.
- **Regression test coverage** added for routing correctness, read-only UI behaviour, internal conversation filtering, and workspace cleanup.

## Validation

```shell
dotnet test BotNexus.slnx --nologo --tl:off
```


sytone:
@copilot resolve the merge coflicts

PR #337, opened 2026-05-18
fix(mobile): reload history on conversation switch; filter hub events by session

larry-fox-lobster:
## Problems fixed

**1. Conversation switch does not reload history**
`OnConvChanged` set `ActiveConversationId` but never fetched history. Switching conversations left the chat area empty or stale.

Fix: make `OnConvChanged` async, clear messages, reset `ActiveSessionId`, await `LoadHistoryAsync`.

**2. Hub events bleed across conversations/agents**
Streaming events for a different session (slow response, background tool call) were appended to whatever view was active.

Fix: `IsActiveSession(e)` guard in all hub handlers — skip events where `e.SessionId != State.ActiveSessionId`. Null on either side = pass-through (safe before first send).

### DIFF
--- a/src/extensions/BotNexus.Extensions.Channels.SignalR.BlazorClient.Mobile/Pages/Chat.razor
+++ b/src/extensions/BotNexus.Extensions.Channels.SignalR.BlazorClient.Mobile/Pages/Chat.razor
@@ -163,9 +163,22 @@ else
         }
     }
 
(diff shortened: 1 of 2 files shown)


larry-fox-lobster:
Superseded by #338 which replaces MobileGatewayClient entirely with Core services. The session filtering and conversation switch reload are now handled correctly by GatewayEventHandler and AgentInteractionService from Core.

PR #396, opened 2026-05-20
fix(conversations): update update date and notify clients on message activity

### LINKED ISSUE #382: [Portal] Conversation timestamp and sort order don't update when messages are exchanged
# Bug: Conversation timestamp and sort order do not update when messages are exchanged

## Summary

The `UpdatedAt` timestamp shown in the conversation list does not update when an agent turn completes, and conversations are not reordered to reflect recent activity. The timestamp only updates when a conversation is explicitly mutated via the REST API (create, rename, archive). Normal agent message activity — the most common "update" — does not trigger any notification to the client.

### Link 1: `DefaultConversationRouter` never calls `IConversationChangeNotifier`

`DefaultConversationRouter` correctly sets `conversation.UpdatedAt = DateTimeOffset.UtcNow` and saves to the store when a session is resolved or bound to a conversation. However it **never calls `IConversationChangeNotifier.NotifyConversationChangedAsync`**.

The notifier is only called from `ConversationsController` for explicit REST mutations:

```csharp
// Only these three places call NotifyConversationChangedAsync:
// ConversationsController.CreateAsync
// ConversationsController.UpdateAsync  
// ConversationsController.ArchiveAsync
```

`DefaultConversationRouter.RouteAsync` — which runs on every inbound message — updates `UpdatedAt` in the DB but emits no SignalR notification. The client never learns the timestamp changed.

### Link 2: `HandleMessageEnd` in `GatewayEventHandler` does not update `ConversationState.UpdatedAt` client-side

Even if a SignalR notification were firing, the client-side `ConversationState.UpdatedAt` field is only populated from the REST API response (initial load + `HandleConversationChanged` refresh). The `HandleMessageEnd` handler adds the message to `conv.Messages` but never touches `conv.UpdatedAt`.

So even without a server push, the client could update the timestamp locally on message receipt — it doesn't.

### Result

```
Agent turn completes
  → GatewayEventHandler.HandleMessageEnd fires
    → conv.Messages gets new message ✓
    → conv.UpdatedAt unchanged ✗
    → conversation list order unchanged ✗
  → DefaultConversationRouter sets conversation.UpdatedAt in DB ✓
    → IConversationChangeNotifier NOT called ✗
    → client never notified ✗
```

The timestamp shown in the sidebar is frozen at the time the conversation was created or last renamed.

## Where the UI Renders It

**Desktop sidebar** (`ConversationsPanel.razor` or equivalent):
```
conversations.OrderByDescending(c => c.UpdatedAt)
```

**Mobile** (`Chat.razor`):
```csharp
.Where(c => c.Status == "Active")
.OrderByDescending(c => c.UpdatedAt)
```

Both are already ordering by `UpdatedAt` — the data just never arrives.

### Fix 1 — Server: call the notifier from `DefaultConversationRouter`

After setting `conversation.UpdatedAt` and saving:

```csharp
await _conversationStore.SaveAsync(conversation, ct);

// Notify clients that this conversation was updated
if (_changeNotifier is not null)
    await _changeNotifier.NotifyConversationChangedAsync("updated", agentId, conversation.ConversationId, ct);
```

`IConversationChangeNotifier` should be injected into `DefaultConversationRouter` (it's already registered as `SignalRConversationChangeNotifier`).

This is the correct long-term fix — the server owns the source of truth.

### Fix 2 — Client: update `ConversationState.UpdatedAt` in `HandleMessageEnd`

As a defensive client-side fix regardless of server push:

```csharp
public void HandleMessageEnd(AgentStreamEvent evt)
{
    // ... existing message append logic ...

    // Update timestamp so sort order reflects activity
    conv.UpdatedAt = DateTimeOffset.UtcNow;

    _store.NotifyChanged();
}
```

This ensures the list re-sorts immediately on streaming completion even before the server `ConversationChanged` event arrives.

### Fix 3 — `ConversationChangedPayload` should carry the new `UpdatedAt`

Currently `ConversationChangedPayload` only carries `changeType`, `agentId`, `conversationId` — the client has to do a full refresh API call to learn the new timestamp. The payload should include `updatedAt` so the client can update in-place without a round-trip:

```csharp
public record ConversationChangedPayload(
    string ChangeType,
    string AgentId,
    string ConversationId,
    DateTimeOffset? UpdatedAt = null);  // add this
```

And `HandleConversationChanged` can apply it directly instead of triggering `RefreshConversationsAsync` on every notification.

## Acceptance Criteria

- [ ] After an agent turn completes, the conversation's `UpdatedAt` timestamp updates in the sidebar
- [ ] The conversation list re-sorts to show the most recently active conversation first
- [ ] `DefaultConversationRouter` calls `IConversationChangeNotifier` after updating `UpdatedAt`
- [ ] `HandleMessageEnd` updates `ConversationState.UpdatedAt` client-side
- [ ] `ConversationChangedPayload` carries `updatedAt` to avoid unnecessary REST round-trips
- [ ] Covered by unit tests for `DefaultConversationRouter` notification path

sytone:
Closes #382

## Summary

Fixes the frozen conversation timestamp and broken sort order in the portal sidebar. Three coordinated fixes:

### Fix 1 — Server: `DefaultConversationRouter` calls `IConversationChangeNotifier`

`DefaultConversationRouter` now accepts an optional `IConversationChangeNotifier` (injected automatically via DI). After any `SaveAsync` that sets `UpdatedAt` (new conversation, session change, reopen/reactivate), it calls `NotifyConversationChangedAsync` so clients receive a `ConversationChanged` SignalR push.

### Fix 2 — Client: `HandleMessageEnd` updates `conv.UpdatedAt`

As a defensive client-side fix, `GatewayEventHandler.HandleMessageEnd` now sets `conv.UpdatedAt = DateTimeOffset.UtcNow` immediately on streaming completion. This keeps the sort order current even before the server push arrives.

### Fix 3 — `ConversationChangedPayload` carries `UpdatedAt`

`ConversationChangedPayload` in both server (`BotNexus.Extensions.Channels.SignalR`) and client (`BlazorClient.Core`) HubContracts now includes `DateTimeOffset? UpdatedAt`. `SignalRConversationChangeNotifier` passes `DateTimeOffset.UtcNow`. `HandleConversationChanged` fast-paths when `UpdatedAt` is present — applies it directly to the local conversation state without a REST round-trip; falls back to full refresh for create/archive events.

## Tests

3 new unit tests in `DefaultConversationRouterTests`:
- `ResolveInbound_NewConversation_CallsNotifierWithUpdated` — notifier fires on new conversation
- `ResolveInbound_ExistingConversationSameSession_DoesNotCallNotifier` — no spurious fires when nothing changed
- `ResolveInbound_WithoutNotifier_DoesNotThrow` — backwards-compatible null notifier

### DIFF
--- a/src/extensions/BotNexus.Extensions.Channels.SignalR.BlazorClient.Core/Services/GatewayEventHandler.cs
+++ b/src/extensions/BotNexus.Extensions.Channels.SignalR.BlazorClient.Core/Services/GatewayEventHandler.cs
@@ -1,4 +1,4 @@
-﻿using System.Diagnostics.CodeAnalysis;
+using System.Diagnostics.CodeAnalysis;
 using System.Text.Encodings.Web;
 using System.Text.Json;
 
@@ -250,6 +250,9 @@ public sealed class GatewayEventHandler : IGatewayEventHandler, IDisposable
         conv.StreamState.IsStreaming = false;
         agent.IsStreaming = false;
         agent.ProcessingStage = null;
+
+        // Update timestamp client-side so sort order reflects recent activity (issue #382).
+        conv.UpdatedAt = DateTimeOffset.UtcNow;
 
         if (agent.AgentId != _store.ActiveAgentId)
             agent.UnreadCount++;
@@ -556,10 +559,25 @@ public sealed class GatewayEventHandler : IGatewayEventHandler, IDisposable
     public void HandleConversationChanged(ConversationChangedPayload payload)
     {
         // Server notified that a conversation was created, updated, or archived.
-        // Re-fetch conversation list for the affected agent so the sidebar stays current.
         var agentId = payload.AgentId;
         if (string.IsNullOrWhiteSpace(agentId)) return;
 
+        // Fast path: if the payload carries the new UpdatedAt, apply it directly
+        // to the client state without a REST round-trip (issue #382).
+        if (payload.UpdatedAt.HasValue && _store.GetAgent(agentId) is { } agent)
+        {
+            var convId = payload.ConversationId;
+            if (!string.IsNullOrWhiteSpace(convId) &&
+                agent.Conversations.TryGetValue(convId, out var conv))
+            {
+                conv.UpdatedAt = payload.UpdatedAt.Value;
+                _store.NotifyChanged();
+                return;
+            }
+        }
+
+        // Fallback: full conversation list refresh (covers create/archive and cases
+        // where the conversation is not yet in the local store).
         _ = RefreshConversationsAsync(agentId);
     }
 
@@ -591,3 +609,4 @@ public sealed class GatewayEventHandler : IGatewayEventHandler, IDisposable
         _hub.OnDisconnected -= HandleDisconnected;
     }
 }
+
--- a/src/extensions/BotNexus.Extensions.Channels.SignalR.BlazorClient.Core/Services/HubContracts.cs
+++ b/src/extensions/BotNexus.Extensions.Channels.SignalR.BlazorClient.Core/Services/HubContracts.cs
@@ -1,4 +1,4 @@
-﻿using System.Text.Json.Serialization;
+using System.Text.Json.Serialization;
 
 namespace BotNexus.Extensions.Channels.SignalR.BlazorClient.Services;
 
@@ -148,7 +148,8 @@ public sealed record AgentsChangedPayload(
 public sealed record ConversationChangedPayload(
     [property: JsonPropertyName("changeType")] string ChangeType,
     [property: JsonPropertyName("agentId")] string AgentId,
-    [property: JsonPropertyName("conversationId")] string ConversationId);
+    [property: JsonPropertyName("conversationId")] string ConversationId,
+    [property: JsonPropertyName("updatedAt")] DateTimeOffset? UpdatedAt = null);
 
 /// <summary>Kind of steering feedback event.</summary>
 public enum SteeringFeedbackKind
@@ -164,3 +165,5 @@ public sealed record SteeringFeedbackPayload(
     [property: JsonPropertyName("agentId")] string AgentId,
     [property: JsonPropertyName("sessionId")] string SessionId,
     [property: JsonPropertyName("kind")] SteeringFeedbackKind Kind);
+
+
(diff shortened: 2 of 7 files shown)


sytone (commented):

Solid three-layered fix for #382.

**Strengths:**
- Server-side notifier injection is backwards-compatible (optional DI parameter + null guard)
- Fast-path `UpdatedAt` propagation avoids unnecessary REST round-trips; fallback to full refresh is correct for create/archive events
- Client-side defensive update in `HandleMessageEnd` is a good belt-and-suspenders approach
- 3 focused unit tests cover new-conversation fires, idempotent second-resolve doesn't fire, and null-notifier backwards compat

**Minor observations:**
- The trailing blank lines added to several files (HubContracts.cs ×2, .csproj) are cosmetic noise — harmless but worth noting
- `NotifyChangedAsync` fires `"updated"` for both session-change and reactivation paths, which is correct; the client fallback handles create/archive separately

No blockers. Good to merge.

sytone (commented):
CI: all green ✅

- Root cause addressed: `DefaultConversationRouter` never called `IConversationChangeNotifier` after save ✓
- Three coordinated fixes (server notifier, client defensive update, payload carries `UpdatedAt`) are all correct ✓
- Tests cover: new conversation fires notifier, same-session no spurious fire, null notifier backwards-compat ✓
- Fast-path on `UpdatedAt` avoids REST round-trip – good performance consideration ✓
- Server and client HubContracts both updated consistently ✓
- Sad path (no notifier) tested ✓

sytone (commented):
LGTM ✅

Solid fix for #382. Two-layer approach is well-designed: client-side `HandleMessageEnd` stamps `UpdatedAt` immediately for snappy UX, and `DefaultConversationRouter` now fires `IConversationChangeNotifier` with the timestamp when conversations are touched. The fast-path in `HandleConversationChanged` avoids a REST round-trip when the payload already carries the new timestamp.

Tests cover the notifier being called on new/updated conversations, the null-notifier backwards-compat path, and the stable-session no-op case.

sytone (commented):

Prior LGTM reviews already posted. `DefaultConversationRouter` now fires `IConversationChangeNotifier` with `UpdatedAt` payload. Client fast-path avoids REST round-trips. Tests cover notifier called, idempotent no-op, and null-notifier compat. Ready to merge.

PR #397, opened 2026-05-20
fix(portal): refresh workspace and reports panels on agent turn-end

### LINKED ISSUE #345: [Portal] Workspace and Reports panels do not refresh after agent writes new files
## Problem

The Workspace file tree and Reports panel both load their data once on component mount (`OnInitializedAsync`) and never refresh. If an agent writes new files to the workspace or reports directory during a session, they are invisible until the user navigates away and back.

## Root Cause

Neither `WorkspacePanel.razor` nor `ReportsPanel.razor` subscribe to any SignalR events or implement any refresh mechanism. Both call the API exactly once at mount time:

- `WorkspacePanel` calls `LoadRootAsync()` in `OnInitializedAsync` only
- `ReportsPanel` calls `LoadReportsAsync()` in `OnInitializedAsync` only

The API controllers (`WorkspaceController`, `ReportsController`) read from disk live with no caching, so the data is always fresh **if called** — the problem is the UI never calls again.

## Reproduction

1. Open an agent and navigate to the Workspace or Reports tab
2. Ask the agent to generate a report or write a file
3. Agent completes successfully
4. New file does not appear in the UI — stale listing from mount time is shown
5. Navigate away and back — file appears

## Expected Behaviour

The Workspace and Reports panels should reflect the current state of the filesystem without requiring navigation. Options (in preference order):

1. **SignalR push** — Gateway emits a `WorkspaceChanged` / `ReportsChanged` event after agent turn completes; panels subscribe and re-fetch
2. **Refresh on agent turn end** — panels listen for the existing turn-complete SignalR event and reload
3. **Manual refresh button** — user-initiated reload as a minimum viable fix
4. **Polling** — fallback if push is not feasible

Option 2 (reload on turn-end event) is the lowest-effort correct fix given SignalR is already wired into the portal. Option 1 is the ideal long-term solution.

## Affected Components

- `WorkspacePanel.razor`
- `ReportsPanel.razor`
- Potentially `WorkspaceFileTree.razor` (cached directory entries dictionary also never invalidated)

sytone:
Closes #345

## Problem

`WorkspacePanel` and `ReportsPanel` both loaded their data exactly once at `OnInitializedAsync` mount time. If an agent wrote new files to the workspace or reports directory during a session, they were invisible until the user navigated away and back.

## Root Cause

Neither panel subscribed to any events or implemented any refresh mechanism. Both called their API exactly once at mount.

### `WorkspacePanel.razor`
- Added optional `Store` `[Parameter]` (`IClientStateStore?`)
- Subscribe to `Store.OnChanged` in `OnInitializedAsync`; unsubscribe in `Dispose`
- Track `_wasStreaming` for the panel's `AgentId`
- When streaming transitions `true → false` for this agent: clear all cached directory responses and reload the root
- Background agent state changes (different agent) do not trigger a reload

### `ReportsPanel.razor`
- Same pattern: optional `Store` `[Parameter]`, subscribe/unsubscribe, `_wasStreaming` guard, reload `LoadReportsAsync()` on turn-end

### `AgentPanel.razor`
- Passes its existing `Store` to both child panels

## Tests (4 new, TDD)

- `WorkspacePanelTests.Refreshes_workspace_when_active_agent_turn_ends`
- `WorkspacePanelTests.Does_not_refresh_workspace_when_background_agent_turn_ends`
- `ReportsPanelTests.Refreshes_reports_when_active_agent_turn_ends`
- `ReportsPanelTests.Does_not_refresh_reports_when_background_agent_turn_ends`

All 285 BlazorClient tests pass. Full suite: 0 failures.

## Design Notes

- `Store` is a `[Parameter]` (not `@inject`) so existing tests that don't provide a store continue to work without modification — fully backwards compatible
- The directory cache is cleared on reload so expanded/cached entries reflect the new filesystem state
- `InvokeAsync` is used for thread-safe Blazor component updates from the event callback

### DIFF
--- a/src/extensions/BotNexus.Extensions.Channels.SignalR.BlazorClient/Components/AgentPanel.razor
+++ b/src/extensions/BotNexus.Extensions.Channels.SignalR.BlazorClient/Components/AgentPanel.razor
@@ -1,4 +1,4 @@
-@inject IClientStateStore Store
+﻿@inject IClientStateStore Store
 @inject NavigationManager Nav
 @implements IDisposable
 
@@ -40,7 +40,7 @@
         <section id="@GetPanelId(AgentPanelTab.Workspace)"
                  class="agent-tab-pane @(IsActiveTab(AgentPanelTab.Workspace) ? "active" : "")"
                  role="tabpanel">
-            <WorkspacePanel AgentId="@AgentId" />
+            <WorkspacePanel AgentId="@AgentId" Store="Store" />
         </section>
         }
 
@@ -49,7 +49,7 @@
         <section id="@GetPanelId(AgentPanelTab.Reports)"
                  class="agent-tab-pane @(IsActiveTab(AgentPanelTab.Reports) ? "active" : "")"
                  role="tabpanel">
-            <ReportsPanel AgentId="@AgentId" />
+            <ReportsPanel AgentId="@AgentId" Store="Store" />
         </section>
         }
 
--- a/src/extensions/BotNexus.Extensions.Channels.SignalR.BlazorClient/Components/ReportsPanel.razor
+++ b/src/extensions/BotNexus.Extensions.Channels.SignalR.BlazorClient/Components/ReportsPanel.razor
@@ -1,4 +1,4 @@
-@inject IGatewayRestClient GatewayRestClient
+﻿@inject IGatewayRestClient GatewayRestClient
 @inject IJSRuntime JS
 @using Microsoft.AspNetCore.Components
 @implements IDisposable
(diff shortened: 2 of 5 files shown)


sytone (commented):
CI: all green ✅

- Happy paths: workspace refreshes on active-agent turn end, reports refreshes on active-agent turn end ✓
- Sad paths: background-agent turn end does NOT trigger refresh for either panel ✓
- `Store` as optional `[Parameter]` (not injected) preserves backwards compatibility with existing tests ✓
- Directory cache cleared on reload – ensures no stale expanded entries ✓
- `InvokeAsync` used for thread-safe component updates from event callback ✓
- Subscribe/unsubscribe pattern with `IDisposable` prevents memory leaks ✓
- `AgentPanel.razor` passes its existing `Store` to children – minimal wiring change ✓

sytone (commented):
LGTM ✅

Good fix for #345. The `_wasStreaming → !isStreaming` edge-detection pattern is clean and only triggers a reload at turn-end. State subscription is properly torn down in `Dispose`. Tests verify both the happy path (reload fires on turn-end) and the guard (background agent turn-end does NOT reload the active panel).

sytone (commented):

Prior LGTM reviews already posted. Turn-end triggered workspace/reports panel refresh is clean. Subscribe/dispose pattern is correct. Tests confirm refresh fires only for the active agent. Ready to merge.

PR #399, opened 2026-05-20
fix(conversation): trigger agent turn when action=new is called with a message

### LINKED ISSUE #285: [Gateway] New conversation with initial message not triggering agent turn
## Summary
When a new conversation is created with an initial user message via the conversation tool or API, the message appears in the channel but the agent never responds. The message is created but not acted upon.

## Steps to Reproduce
1. Call `conversation(action='new', message='Hello')`
2. Observe the new conversation in the portal
3. The user message is visible but the agent produces no response

## Expected Behavior
Agent responds to the initial message as a normal user turn.

## Actual Behavior
Message is stored, no agent turn triggered.

## Likely Cause
The session is not fully initialized before the message is injected, causing it to be silently dropped. Or the message is not being routed into the agent's session loop after conversation creation.

## Related
- #252 (conversation tool new action should support initial message)
- #253 (message injection timing)
- #228 (message queue injection)

sytone:
Closes #285

## Summary

When `conversation(action='new', message='...')` was called, the initial user message was stored in session history but the agent never produced a response. The message sat silently in the session; no agent turn was triggered.

## Root Cause

`ConversationTool.NewAsync` created the session, added the user message, and saved everything — but never dispatched to the gateway pipeline (`IChannelDispatcher`). Without a dispatch, the session queue never picked up the work.

### `ConversationTool`
- Added optional `IChannelDispatcher? channelDispatcher` parameter (backwards-compatible — defaults to null)
- After saving the seeded session, dispatches a synthetic `InboundMessage` with `ChannelType=internal`, `TargetAgentId`, `SessionId`, and `ConversationId` set correctly
- Gateway session queue receives the message and runs the agent turn as normal
- When `channelDispatcher` is null (no DI registration), existing behaviour is preserved: history seeded, no turn triggered — fully backwards-compatible

### `InProcessIsolationStrategy`
- Resolves `IChannelDispatcher` from DI and passes it to `ConversationTool`
- Added `using BotNexus.Gateway.Abstractions.Channels;`

## Tests (TDD — 3 new tests)

- `New_WithMessage_AndDispatcher_DispatchesInboundMessageToTriggerAgentTurn` — happy path: verifies `DispatchAsync` called with correct `TargetAgentId`, `Content`, `SessionId`, `ConversationId`, `ChannelType=internal`
- `New_WithMessage_WithoutDispatcher_DoesNotThrowAndStillSeedsHistory` — sad path / backwards-compat: no dispatcher, no throw, history still seeded
- `New_WithoutMessage_DoesNotInvokeDispatcher` — edge case: no message → dispatcher never called

All 1,524 gateway tests pass. Full suite: 0 failures.

### DIFF
--- a/src/gateway/BotNexus.Gateway/Isolation/InProcessIsolationStrategy.cs
+++ b/src/gateway/BotNexus.Gateway/Isolation/InProcessIsolationStrategy.cs
@@ -16,6 +16,7 @@ using BotNexus.Gateway.Abstractions.Hooks;
 using BotNexus.Gateway.Abstractions.Isolation;
 using BotNexus.Gateway.Abstractions.Models;
 using BotNexus.Gateway.Abstractions.Security;
+using BotNexus.Gateway.Abstractions.Channels;
 using BotNexus.Gateway.Abstractions.Sessions;
 using BotNexus.Domain.Primitives;
 using BotNexus.Gateway.Agents;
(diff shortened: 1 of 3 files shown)


sytone (commented):
LGTM ✅

Clean fix for #285. Dispatching a synthetic `InboundMessage` through `IChannelDispatcher` is the right approach — it reuses the normal inbound path rather than introducing a side-channel. The three tests cover all meaningful paths:
- Dispatcher invoked with correct fields on `action=new` + message
- Null dispatcher is a no-op (backwards compat)
- No message → no dispatch

sytone (commented):

Prior LGTM review already posted. Synthetic `InboundMessage` dispatch through `IChannelDispatcher` is the correct approach for triggering agent turns on `action=new`. Three tests cover all paths. Ready to merge.

PR #408, opened 2026-05-20
fix(gateway): preserve compaction summary when agent handle is recreated

### LINKED ISSUE #402: [Gateway] Pinned context files (SOUL.md, MEMORY.md, etc.) may be dropped during session compaction
## Summary

When `/compact` prunes session history, pinned context files injected at session bootstrap (e.g. `SOUL.md`, `MEMORY.md`, `WORLD.md`, project context files) are not re-injected into the compacted history. The next turn after compaction runs without those foundational context files.

## Root Cause

OpenClaw audit (SHA `ad925bd4`): upstream found that policy files (AGENTS.md equivalent) were dropped during bootstrap truncation. BotNexus `LlmSessionCompactor` replaces `session.History` with only the summary + recent messages — it does not re-pin workspace context files into the compacted history.

## Impact

- Agent loses persona, memory, and world context after compaction
- Behavior drift post-compaction — agent may not follow SOUL.md tone/guardrails
- Silent: no error is thrown, degraded behavior is the only signal

## Proposed Fix

- `LlmSessionCompactor` should accept a list of `PinnedContextFiles` (from agent config)
- After compaction, re-inject pinned file content as system messages at the head of the compacted history
- OR: mark pinned context entries in history with a flag so compactor preserves them

## Acceptance Criteria

- [ ] After `/compact`, SOUL.md / MEMORY.md content is still present in the active session context
- [ ] Compacted history token count is still reduced (pinned files don't cancel the savings)
- [ ] Test: compact a session with pinned files, verify pinned content survives

sytone:
Closes #402

### Root cause
`InProcessIsolationStrategy.CreateAsync` filtered history entries to `User` and `Assistant` roles only when seeding initial messages for a new agent handle. `IsCompactionSummary` entries have `Role = System` and were excluded. After a gateway restart (new handle creation), the compaction summary was silently dropped and the LLM lost summarised prior context — exactly the bug described.

### Fix
- `InProcessIsolationStrategy.CreateAsync`: extend the history filter to also include entries where `IsCompactionSummary == true`, converting them to `SystemAgentMessage`. Regular system entries (`IsCompactionSummary = false`) remain excluded to avoid duplicating the system prompt.

### Files changed
- `src/gateway/BotNexus.Gateway/Isolation/InProcessIsolationStrategy.cs` — updated filter predicate
- `tests/gateway/BotNexus.Gateway.Tests/InProcessIsolationStrategyTests.cs` — 3 new tests

## Tests
- `CreateAsync_WithCompactionSummaryInHistory_InjectsSummaryAsSystemMessage` — happy path: summary + turns → all 3 messages present in correct order
- `CreateAsync_WithOnlyCompactionSummary_InjectsSummaryOnly` — edge case: summary with no preserved turns still injects
- `CreateAsync_WithRegularSystemEntryInHistory_ExcludesNonSummarySystemMessages` — regression guard: plain system entries still excluded

All 1539/1539 gateway tests pass.

### DIFF
--- a/src/gateway/BotNexus.Gateway/Isolation/InProcessIsolationStrategy.cs
+++ b/src/gateway/BotNexus.Gateway/Isolation/InProcessIsolationStrategy.cs
@@ -287,13 +287,16 @@ public sealed class InProcessIsolationStrategy : IIsolationStrategy
         List<AgentMessage>? initialMessages = null;
         if (context.History.Count > 0)
         {
-            // Only inject user and assistant messages from history. Tool-role entries
-            // become orphaned ToolResultMessages (no matching tool_use in the preceding
-            // assistant message) which causes the LLM provider to reject the conversation.
-            // System entries are also excluded — the agent's system prompt is set separately.
+            // Inject compaction summary entries as system messages so the LLM retains
+            // summarised context when the handle is recreated (e.g. after a gateway restart).
+            // Tool-role entries are excluded: they become orphaned ToolResultMessages
+            // (no matching tool_use in the preceding assistant message) which causes
+            // LLM provider rejection. Regular system entries (IsCompactionSummary=false)
+            // are also excluded — the agent's system prompt is set separately.
             initialMessages = context.History
                 .Where(e => e.Role == Domain.Primitives.MessageRole.User
-                         || e.Role == Domain.Primitives.MessageRole.Assistant)
+                         || e.Role == Domain.Primitives.MessageRole.Assistant
+                         || (e.Role == Domain.Primitives.MessageRole.System && e.IsCompactionSummary))
                 .Select(ConvertSessionEntryToAgentMessage)
                 .ToList();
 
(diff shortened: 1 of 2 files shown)


sytone (commented):

**Title check:** `fix(gateway): preserve compaction summary when agent handle is recreated` — ✅ correct conventional commit format, no issue number in title.

**Summary:** Fixes #402 — compaction summary entries (`IsCompactionSummary == true`, `Role = System`) were silently dropped when `InProcessIsolationStrategy.CreateAsync` filtered history to `User`/`Assistant` only. Fix extends the predicate to include summary entries and converts them to `SystemAgentMessage`.

**Strengths:**
- Root cause is precisely identified and the fix is surgical — one predicate change, no other production surface touched
- Regular system entries remain excluded to avoid duplicating the system prompt — correct discrimination
- Three tests cover all meaningful paths: summary + turns (happy path), summary only (edge case), plain system excluded (regression guard)
- 1539/1539 gateway tests passing confirms no regressions

- Converting summary entries to `SystemAgentMessage` is correct but worth confirming the model provider interprets a system-role message mid-history correctly (most modern providers handle this fine)
- No changes to `IsolationStrategy` interface or other strategies — consider whether `ExternalIsolationStrategy` has the same gap if it exists

**Verdict:** ✅ LGTM — CI green, targeted fix, comprehensive tests. Ready to merge.

PR #423, opened 2026-05-20
fix(gateway): fan-out stream events to SignalR observers on cross-channel messages

### LINKED ISSUE #332: [Portal] Web UI doesn't live-update when messages arrive via external channels (Telegram, etc.)
## Bug

When a message is sent to an agent via Telegram (or any non-SignalR channel), the web portal does not update in real-time. The user must manually reload the page to see the new messages.

## Root Cause

The channel architecture is single-path: when a message arrives via Telegram, it's processed and the response is sent back through the Telegram `IChannelAdapter` only. The SignalR adapter is never notified, so no `MessageStart`/`ContentDelta`/`MessageEnd` events are broadcast to connected web clients.

## Evidence from logs (this session)

```
19:40:40 — Telegram message received, channel resolved as type=telegram
19:40:54 — Response sent to Telegram sendMessage API
19:41:10 — User manually reloads web portal (SignalR disconnect/reconnect)
```

No SignalR broadcast events occurred between the Telegram message arrival and the manual page reload. The web UI was connected and subscribed to the session group the entire time.

## Expected Behaviour

When a message arrives on **any** channel and the conversation/session is open in the web UI, the web UI should update live. The SignalR adapter should receive cross-channel notifications for:
- Incoming user messages (so the chat shows the user's message immediately)
- Assistant responses (streaming deltas)
- Tool calls and results

## Suggested Approach

After the primary channel processes a message, broadcast a notification to the SignalR session group so any connected web clients update. This could be:
1. **Observer/event pattern** — the agent execution pipeline emits events that all interested channel adapters can subscribe to
2. **Explicit cross-channel broadcast** — after responding on the primary channel, also push to SignalR if clients are connected
3. **Shared event bus** — decouple channel notification from channel routing

Option 1 is cleanest long-term — the pipeline already has stream events, they just need to fan out to all adapters with active listeners, not just the originating channel.

## Impact

Major UX issue — if a user is chatting via Telegram and has the web portal open (common during development/monitoring), the portal looks stale and unresponsive.

sytone:
Closes #332

## Root Cause

When an agent processes a message arriving via a non-SignalR channel (e.g. Telegram), stream events (`MessageStart`, `ContentDelta`, `MessageEnd`, etc.) were only forwarded to the originating channel adapter. SignalR bindings on the same conversation received no notification until `FanOutResponseAsync` ran at the end of the turn — delivering only the final assembled message, not the live stream.

This caused the web portal to appear stale/unresponsive when users were active on Telegram while monitoring via the portal.

### `GatewayHost.cs`
- Before entering the streaming loop for a non-SignalR originating message, resolve any SignalR bindings on the conversation via `IConversationRouter.GetOutboundBindingsAsync`
- Inside `OnEventAsync`, after delivering each stream event to the primary channel, fan-out the same event to all resolved SignalR observer bindings
- Guard: skip the extra resolution when the originating channel is already SignalR (avoids unnecessary overhead and double-send)
- Best-effort: per-observer exceptions are caught/logged and do not interrupt the primary stream

### `GatewayHostTests.cs` (+2 tests)
- `StreamEvents_FanOutToSignalRObserverBindings_WhenOriginatingChannelIsNotSignalR` — happy path: Telegram message → SignalR observer receives all 3 stream events (MessageStart, ContentDelta, MessageEnd)
- `StreamEvents_DoNotFanOutToSignalRObserver_WhenOriginatingChannelIsSignalR` — sad path: SignalR originator → no observer doubling, stream events delivered exactly once via primary path

## Testing

1550/1550 tests pass (1548 pre-existing + 2 new).


sytone:

**Title check:** `fix(gateway): fan-out stream events to SignalR observers on cross-channel messages` — ✅ correct conventional commit format, no issue number in title.

**Merge conflict status:** ✅ clean (MERGEABLE)

**Summary:** Fixes #332 — stream events (MessageStart, ContentDelta, MessageEnd) now fan out to SignalR observer bindings when the originating channel is non-SignalR (e.g. Telegram). Web portal no longer appears stale during Telegram-originated turns.

**Strengths:**
- Pre-resolving observer bindings *before* the streaming loop is the correct approach — avoids per-event binding lookup overhead
- Originator guard (`isSignalROriginator` short-circuit) correctly prevents double-send on SignalR-originated turns
- Per-observer try/catch ensures a failing observer never interrupts the primary stream
- 2 new tests (happy + sad path) covering the exact fan-out and no-doubling scenarios
- 1550/1550 pass

- Binding resolution happens once before the loop — if a new SignalR session connects mid-stream, it won't receive the in-progress stream events. This is an acceptable snapshot-at-start trade-off.
- The `IConversationRouter.GetOutboundBindingsAsync` call adds one DB/lookup per cross-channel turn — negligible, but worth noting if high throughput is ever a concern.

**Verdict:** ✅ LGTM — CI green, clean merge, targeted fix with good test coverage. Ready to merge.

sytone:
**Merge conflict resolved** — main's `UserInputRequired` handling in the `OnEventAsync` lambda has been incorporated alongside the branch's SignalR observer fan-out loop. Both code paths now coexist: `UserInputRequired` dispatches first (early return), then `IStreamEventChannelAdapter` handles stream events, followed by the observer fan-out. Build and 1567/1567 tests pass locally.

Issue #363, opened 2026-05-19
[Gateway] Gateway restart during agent run causes silent data loss — no write-ahead or crash sentinel


sytone:
## Summary

If the gateway restarts while an agent turn is in flight, all progress is silently lost. The session store holds the last persisted snapshot — which predates the current run — and the in-flight LLM stream, tool calls, and partial assistant response are dropped entirely with no trace.

## Root Cause

Session entries are only written to the store at the end of a completed run (`AgentEnd`). The user message that triggered the run is added to the in-memory session object but not flushed to the store before the LLM call begins. There is no crash sentinel or write-ahead record.

On restart:
- `DefaultAgentSupervisor` loses all in-memory agent handles
- The session store has the pre-run state
- The client reconnects and sees a conversation where the user message (and all subsequent activity) never happened

## Impact

- Silent data loss on any gateway restart, OOM kill, or unhandled exception during a run
- User has no indication a run was in progress — the message they sent appears to have vanished
- No recovery path: the partial run cannot be resumed or inspected

## Fix — Write-Ahead + Crash Sentinel

Two layers needed:

**Layer 1 — Write-ahead persistence**
Persist the user message to the session store *before* starting the LLM stream. Currently the message is added to the in-memory `GatewaySession` but `SaveAsync` is not called until after the full run. Moving this save earlier means at minimum the user's input survives a restart.

Relevant location: `GatewayHost.cs` — user message is added to session around line 400, but `ProcessAndSaveAsync` (which calls `SaveAsync`) is not invoked until the stream completes.

**Layer 2 — Crash sentinel / tombstone**
Before starting the LLM call, write a sentinel session entry:
```
role: system, content: "[agent turn in progress — interrupted if visible]"
```
On clean `AgentEnd`, remove or replace it with the real final state. On restart, the sentinel is persisted and visible — the UI can surface "this run was interrupted" rather than a silent gap.

This mirrors the OpenClaw pattern (SHA `1b82c0e3`) where a user-message latch and an error stub are written before the LLM call, preventing retry paths from re-persisting already-saved messages and leaving a recoverable trace on failure.

## Relationship

- #362 — in-flight entries not visible to second client (mid-run visibility fix, TurnEnd flush)
- This issue is the durability / restart-safety complement to #362
- OpenClaw audit 2026-05-18: SHA `1b82c0e3` — persistence latch pattern, same root class

sytone:
PR opened: https://github.com/sytone/botnexus/pull/424

PR #431, opened 2026-05-21
fix(portal): eliminate steering toggle flicker between tool calls

### LINKED ISSUE #253: [Portal][Gateway] Steering toggle flickers mid-turn -- agent state published at too fine granularity
## Bug

Between tool calls and agent messages arriving in the portal chat panel, the steering/direct-message toggle flickers — it switches between "Steer" and direct input mode as each message arrives, creating a confusing and broken UX.

## Expected Behavior

- **Steering should be the default** whenever an agent is mid-turn (processing, calling tools, generating a response).
- User messages sent while an agent is mid-turn should be **held** and injected at the correct point: before the next LLM call, or after the current tool call completes.
- The toggle should only switch to "direct input" mode when the agent is genuinely idle and ready to receive a new user message directly.

## Actual Behavior

The steering/direct toggle oscillates as individual messages and tool call events arrive over SignalR. Each message event appears to update agent state in a way that causes the UI to flip the input mode, even though the agent is still mid-turn.

## Root Cause (Suspected)

The core agent likely has injection point logic set up correctly, but there appears to be a state publication issue — agent availability/execution state is being emitted at too fine a granularity (per-message or per-tool-event), causing the portal to re-evaluate input mode on each event rather than only on genuine turn-boundary transitions.

## Related Issues

- #228 — Message queue injection timing (broader; this is specifically the UI flickering symptom and state publication granularity)
- #194 — Clarify agent availability vs execution state

## Impact

Confusing UX. Users may attempt to send a direct message during a mid-turn window and have it dropped or misrouted.

sytone:
## Summary

Fixes #253 — steering/direct-message toggle flickered between tool calls.

## Root Cause

Between `MessageEnd` and the next `MessageStart`, `conv.StreamState.IsStreaming` is `false` even though the agent turn is still active (tools are executing). `ChatPanel.IsStreaming` was bound directly to `IsStreaming`, so the portal briefly showed the Send button instead of Steer/Abort during that gap.

## Fix

Added `ConversationStreamState.IsTurnActive` computed property:

```csharp
public bool IsTurnActive => IsStreaming || ActiveToolCalls.Count > 0;
```

Updated `ChatPanel.IsStreaming` to use `IsTurnActive` so Steer/Abort controls remain visible for the full duration of a multi-tool turn — not just during LLM streaming.

## Tests

4 new tests in `GatewayEventHandlerTests`:
- `IsTurnActive_is_true_when_streaming`
- `IsTurnActive_is_false_when_not_streaming_and_no_tool_calls`
- `IsTurnActive_is_true_after_message_end_when_tool_calls_still_pending` — the key regression test
- `IsTurnActive_is_false_after_tool_end_and_message_end_both_complete`

All 1579 tests pass.

Closes #253

### DIFF
--- a/src/extensions/BotNexus.Extensions.Channels.SignalR.BlazorClient.Core/Services/Abstractions/IClientStateStore.cs
+++ b/src/extensions/BotNexus.Extensions.Channels.SignalR.BlazorClient.Core/Services/Abstractions/IClientStateStore.cs
@@ -1,4 +1,4 @@
-namespace BotNexus.Extensions.Channels.SignalR.BlazorClient.Services;
+﻿namespace BotNexus.Extensions.Channels.SignalR.BlazorClient.Services;
 
 /// <summary>
 /// Single in-memory source of truth for all portal state.
@@ -244,6 +244,7 @@ public sealed class ConversationState
     public ConversationStreamState StreamState { get; } = new();
 }
 
+
 /// <summary>Stream-buffer state for an active or recently active conversation.</summary>
 public sealed class ConversationStreamState
 {
@@ -258,4 +259,10 @@ public sealed class ConversationStreamState
 
     /// <summary>In-progress tool calls for this conversation keyed by tool-call ID.</summary>
     public Dictionary<string, ActiveToolCall> ActiveToolCalls { get; } = new();
+
+    /// <summary>Whether the agent turn is still active -- either streaming or awaiting tool results.
+    /// Use this instead of IsStreaming to keep Steer/Abort controls visible
+    /// between the end of an LLM generation and the start of the next one while tools run.</summary>
+    public bool IsTurnActive => IsStreaming || ActiveToolCalls.Count > 0;
 }
+
--- a/src/extensions/BotNexus.Extensions.Channels.SignalR.BlazorClient/Components/ChatPanel.razor
+++ b/src/extensions/BotNexus.Extensions.Channels.SignalR.BlazorClient/Components/ChatPanel.razor
@@ -1,4 +1,4 @@
-@inject IJSRuntime JS
+﻿@inject IJSRuntime JS
 @inject IClientStateStore Store
 @inject IAgentInteractionService Interaction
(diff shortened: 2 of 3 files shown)


sytone:
## Farnsworth Review — #431 fix(portal): eliminate steering toggle flicker between tool calls

**CI:** ✅ All checks passing

**Merge conflicts:** Clean (MERGEABLE)

**Title check:** `fix(portal): eliminate steering toggle flicker between tool calls` — ✅ correct conventional commit format, no issue number in title

**Root cause analysis:** Correct. `IsStreaming` goes false between `MessageEnd` and next `MessageStart` even though the turn is still active due to tool execution. `IsTurnActive` computed property (`IsStreaming || ActiveToolCalls.Count > 0`) captures the full turn lifecycle. Clean fix.

**Test coverage:** 4 new tests in `GatewayEventHandlerTests`:
- `IsTurnActive_is_true_when_streaming` — happy path ✅
- `IsTurnActive_is_false_when_not_streaming_and_no_tool_calls` — sad path ✅
- `IsTurnActive_is_true_after_message_end_when_tool_calls_still_pending` — key regression ✅
- `IsTurnActive_is_false_after_tool_end_and_message_end_both_complete` — clean exit ✅

All happy and sad paths covered. 1579 tests pass.

**Spec completeness vs #253:** `IsTurnActive` replaces `IsStreaming` in `ChatPanel.razor`; `IClientStateStore` interface updated. Fully satisfies the acceptance criteria.

- `IsTurnActive` is computed from `ActiveToolCalls.Count > 0` — if `ToolStart` is received but `ToolEnd` never arrives (e.g. timeout), the portal will show Steer/Abort indefinitely. The timeout/watchdog is a separate concern (#363 covers crash sentinel), but worth noting.
- The interface change to `IClientStateStore` is a non-breaking addition — no consumers affected.

**Verdict:** ✅ LGTM — minimal, targeted fix, exactly right semantics, well-tested. Ready to merge.

PR #460, opened 2026-05-21
fix(portal): defer conversation list refresh while agent is streaming

### LINKED ISSUE #456: [Portal] Stream events stop rendering mid-turn when conversation tool switches active conversation
[Portal] Stream events stop rendering mid-turn when conversation tool switches active conversation

## Summary

When an agent calls `conversation(action=new)` mid-turn, the portal stops receiving SignalR stream events for the originating conversation. The turn continues to completion (all messages and tool calls land in the transcript), but they are invisible in the portal until a page refresh.

## Observed Behavior

1. Agent is streaming in conversation A
2. Agent calls `conversation(action=new)` — a new conversation B is created
3. Portal stops rendering stream events immediately at that point
4. Remaining tool calls and assistant messages complete silently
5. After page refresh, all missing content appears (REST hydration catches up)

## Expected Behavior

The originating conversation's stream should continue uninterrupted regardless of what the agent does with the conversation tool mid-turn. A mid-turn conversation switch should not affect the SignalR fan-out for the session that started the turn.

## Root Cause Hypothesis

`conversation(action=new)` likely updates the active `ConversationId` on the session or SignalR binding. Subsequent `GatewayHost` stream events are then fanned out to the new conversation ID. The portal is still displaying the old conversation and misses those events.

## Reproduction

1. Start a long agent turn with multiple tool calls
2. Have the agent call `conversation(action=new, ...)` mid-turn
3. Observe: stream stops updating in portal
4. Refresh page: missing content appears

## Related

#314 / PR #422 — stale ConversationId on conversation switch
#332 / PR #423 — cross-channel fan-out
#441 / PR #443 — bind-on-first-use on reconnect

sytone:
Closes #456

## Problem

When an agent calls `conversation(action=new)` mid-turn, a `ConversationChangedPayload` arrives for a conversation not yet in local state. The fallback `RefreshConversationsAsync` path runs `SeedConversations` synchronously, which overwrites `ActiveSessionId` on existing conversations from a server snapshot that may be slightly stale. Once `TryResolveConversationBySession` breaks, all subsequent stream events (ContentDelta, ToolEnd, MessageEnd) silently fail to route to the originating conversation. The portal goes dark until a page refresh.

## Fix

- Added `_pendingConversationRefresh: HashSet<string>` to `GatewayEventHandler`
- `HandleConversationChanged` fallback path checks `agent.IsStreaming` before firing `RefreshConversationsAsync`; if streaming, queues the agentId instead
- `HandleMessageEnd` and `HandleError` drain the pending set via `DrainPendingConversationRefreshes(agentId!)` after streaming ends
- 3 new unit tests: deferred during stream, drained on MessageEnd, immediate when not streaming

### DIFF
--- a/src/extensions/BotNexus.Extensions.Channels.SignalR.BlazorClient.Core/Services/GatewayEventHandler.cs
+++ b/src/extensions/BotNexus.Extensions.Channels.SignalR.BlazorClient.Core/Services/GatewayEventHandler.cs
@@ -20,6 +20,11 @@ public sealed class GatewayEventHandler : IGatewayEventHandler, IDisposable
     private readonly GatewayHubConnection _hub;
     private readonly HashSet<string> _streamingWhenDisconnected = new();
 
+    // Conversation refreshes that arrive during a streaming turn are deferred
+    // until HandleMessageEnd/HandleError to avoid overwriting ActiveSessionId
+    // mid-turn (breaks stream event routing -- issue #456).
+    private readonly HashSet<string> _pendingConversationRefresh = new();
+
     public GatewayEventHandler(IClientStateStore store, GatewayHubConnection hub)
     {
         _store = store;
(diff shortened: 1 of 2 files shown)


sytone (commented):
**CI:** ✅ All checks passing
**Merge conflicts:** Clean (CLEAN)
**Conventional commit:** ✅ `fix(portal): defer conversation list refresh while agent is streaming`
**Test coverage:** 3 new tests covering happy path (defer during stream, drain on message-end, drain on error). Sad path (simultaneous multi-agent) not tested but low risk given HashSet design.
**Spec completeness:** #456 fully addressed — root cause (SeedConversations overwriting ActiveSessionId mid-turn) correctly fixed via pending-refresh pattern.

LGTM — ready to merge.

PR #565, opened 2026-05-27
fix(gateway): cancellation must not seal cross-world relay session (#553)

### LINKED ISSUE #553: [Gateway] Cross-world relay: OperationCanceledException should not seal session
**Source:** Multi-model critique sweep on PR #550 (Phase 8 / F-11). Bug-hunt agent (gpt-5.3-codex) flagged LOW-4.

## Problem

In CrossWorldFederationController.RelayAsync, the broad catch (Exception ex) block at the end of the relay flow seals the session on any exception, including OperationCanceledException raised by the caller's cancellation token. This means:

1. Caller cancels mid-flight (e.g. HTTP client times out, sender retry policy aborts).
2. Receiver catches OperationCanceledException, seals the session, returns 500.
3. Sender retries -> hits sealed-session 409 guard -> retry fails permanently.

The intended behaviour is: cancellation should rethrow without sealing, so the sender's retry sees the session still Active and can resume.

This is pre-existing — not introduced by Phase 8 — but flagged here so it gets a tracked fix.

## Proposed remediation

1. Add catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested) { throw; } before the catch-all in RelayAsync.
2. Same pattern in AgentExchangeService.ConverseLocalAsync if it has a similar catch shape.
3. Add a test that fires a relay with a cancelled token, asserts:
   - 499/cancellation result (not 500).
   - Session is still Active (NOT sealed).
   - A subsequent reuse call with the same RemoteSessionId succeeds.

## Related

- PR #550 (Phase 8 / F-11)
- src/gateway/BotNexus.Gateway.Api/Controllers/CrossWorldFederationController.cs (RelayAsync catch block)

sytone:
# #553 — Cancellation must not seal the cross-world relay session

## What was wrong

Caller-initiated cancellation of a cross-world relay used to permanently seal the session via the catch-all `catch (Exception)` that fired on every OCE — including the caller's own. The sealed-session 409 guard in `ResolveSessionAsync` then turned a transient client timeout, browser tab close, or user-driven cancellation into a permanently broken sender-retry contract documented in PR #549.

## What changed

Insert a guarded re-throw BEFORE each existing catch-all so caller-initiated cancellation rethrows without sealing:

```csharp
catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
{
    // #553: caller-initiated cancellation must NOT seal the session.
    throw;
}
catch (Exception ex)
{
    // existing seal-on-error logic, now only sees inner-token cancels
    // and genuine failures.
}
```

The `when` filter is the discriminator: inner-token cancellations (downstream HTTP-client timeouts linked into the supervisor token) MUST still fall through to the catch-all and seal, because they are genuine downstream failures — not callers pulling out.

### Per-site coverage map

| Production site | Behavioural test |
|---|---|
| `CrossWorldFederationController.ExecuteRelayAsync` | `RelayAsync_WhenCallerCancelsDuringPromptAsync_RethrowsOce_AndDoesNotSealSession` |
| `AgentExchangeService.ConverseAsync` (local) | `ConverseAsync_LocalPath_WhenCallerCancelsDuringPromptAsync_RethrowsOce_AndDoesNotSealSession` |
| `AgentExchangeService.ConverseAsync` (cross-world relay-out) | `ConverseAsync_CrossWorldPath_WhenCallerCancelsDuringPromptAsync_RethrowsOce_AndDoesNotSealSession` |

## Acceptance criteria (from #553)

| AC | Pinned by | Verdict |
|---|---|---|
| AC #1 — cancelled relay throws OCE, session stays Active | `RelayAsync_WhenCallerCancels...` + local-path twin | ✅ |
| AC #2 — no error metadata recorded on cancellation | same tests assert `Metadata.ContainsKey("error").ShouldBeFalse()` | ✅ |
| AC #3 — retry with same `RemoteSessionId` succeeds | `RelayAsync_AfterCallerCancellation_SessionIsReusableByRetryWithSameRemoteSessionId` — exercises the real `RelayAsync` → `ResolveSessionAsync` reuse path including the 5-invariant `OwnedByRequester` check from PR #549 | ✅ |

## `when` filter vacuity guard

Removing the `when (cancellationToken.IsCancellationRequested)` filter would catch ALL OCEs — including inner-token cancellations that should still seal. The vacuity guard cancels an INNER `CancellationTokenSource` unrelated to the caller, throws OCE bound to the inner token, and asserts the session ends up `Sealed`. If the filter is weakened to a bare `catch (OperationCanceledException)`, the test fails.

- `RelayAsync_WhenInnerTokenCancels_NotCallerToken_StillSealsSession`
- `ConverseAsync_LocalPath_WhenInnerTokenCancels_NotCallerToken_StillSealsSession`

## Architecture fitness fence

`tests/architecture/BotNexus.Architecture.Tests/CancelNoSealArchitectureTests.cs` (10 tests):

- **Per-catch fence** on both production files: every `catch (Exception)` body containing `GatewaySessionStatus.Sealed` must be preceded by the required OCE-rethrow clause within the prior 1500 chars. The 500 → 1500 lookback window covers the substantial rationale comment carried by the controller's catch.
- **Counter-fence** on both production files: each must contain ≥N Sealed-writing catch-alls (controller ≥1, service ≥2), so a future refactor to a helper like `SealSessionAsync(session)` is forced to either keep the literal token OR update the fence regex — otherwise the per-catch fence would silently pass with zero candidates.
- **Vacuity guards**: synthetic catch-all without OCE guard + synthetic bare-OCE catch without `when` filter both correctly violate.
- **False-positive guards**: canonical post-#553 shape (with and without brace-bearing comments inside the rethrow body) correctly does NOT violate. Body regex permits ONE level of nested braces so explanatory comments like `// see Foo {...}` don't push future authors to delete them.
- **Detector specificity**: `ClearActiveSessionAsync` diagnostic catches (which swallow without sealing) are correctly ignored.

## Trade-offs (read these before requesting changes)

- **`Conversation.ActiveSessionId` is intentionally NOT cleared on caller cancellation.** A fast retry sees the conversation as still in-flight rather than briefly flipping idle → active in the portal. Staleness is bounded by `SessionCleanupService` TTL (default 24h, 5m sweep) and `ActiveSessionId` is documented as a diagnostic, not a correctness contract.
- **Catch-all `SaveAsync` calls pass `CancellationToken.None`** on all three sites — caller-token OCE cannot fire during the seal write, so the race that would leave `Status` mutated in-memory but unflushed is structurally avoided.
- **Cancelled turn's user-input is persisted before `PromptAsync`** — on retry the user message would be appended again. This is a real but bounded duplicate-message UX/transcript concern (NOT a security issue per the security review) and is filed as a deferred follow-up. Solving it requires per-turn idempotency keys, which is out of scope for the cancel-no-seal fix.
- **`ClearActiveSessionAsync` diagnostic catches** in both files are deliberately outside the fence's scope — they swallow without sealing and the fence's `Sealed`-keyed detection correctly skips them.

## Multi-model critique sweep

Run before opening the PR:

- **`security-review` (gpt-5.5)** — no vulnerabilities. DoS bounded by `SessionCleanupService` TTL; authz boundary (5-invariant `OwnedByRequester`) preserved.
- **`code-review` plan-vs-impl (claude-opus-4.7)** — all 3 ACs + `when` filter + rethrow + 3-site coverage + fence non-vacuity pinned. Adopted two SHOULD-CONSIDER findings:
  - Regex loosening to permit brace-bearing comments inside rethrow body.
  - Counter-fence on Sealed-writing catch count, so helper-refactor forces fence update.
- **`rubber-duck` bug-hunt (gpt-5.3-codex)** — no blocking issues. Confirmed lock scope is released by `await using`, token identity is the method parameter (not a wrapped/linked variant), vacuity guard truly pins the discriminator.

## Validation

- `dotnet test BotNexus.slnx --nologo --tl:off --no-build` → all green except the known `BotNexus.Cron.Tests.CronSchedulerTests.Scheduler_MultipleDueJobs_RunConcurrently_NotSerially` timing flake (passes on rerun, unrelated).
- `Gateway.Tests`: 1883 passed / 0 failed / 1 skip — +6 new behavioural tests.
- `Architecture.Tests`: 69 passed / 0 failed — +10 new fence tests (5 from initial draft + 5 from critique adoption).

Closes #553.

### DIFF
--- a/src/gateway/BotNexus.Gateway.Api/Controllers/CrossWorldFederationController.cs
+++ b/src/gateway/BotNexus.Gateway.Api/Controllers/CrossWorldFederationController.cs
@@ -237,6 +237,22 @@ public sealed class CrossWorldFederationController(
                 FinishSummary = exchangeFinished ? finishSummary : null
             });
         }
+        catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
+        {
+            // #553: caller-initiated cancellation must NOT seal the session. Sealing here would
+            // poison the session for any sender retry — the sender's next call would hit the
+            // sealed-session 409 guard in ResolveSessionAsync and the exchange would be
+            // permanently broken by a transient client-side timeout / abort. Rethrow without
+            // touching session.Status; the session remains Active and the sender can retry
+            // with the same RemoteSessionId. Note: ActiveSessionId is intentionally LEFT
+            // pinned so a fast retry sees the conversation as in-flight rather than briefly
+            // flipping to "idle" then back to "active" in the portal.
+            //
+            // Guard is `when (cancellationToken.IsCancellationRequested)` so OCEs raised by
+            // any OTHER token (e.g. an internal timeout linked downstream) still fall through
+            // to the catch-all and seal — those are genuine failures, not caller intent.
+            throw;
+        }
         catch (Exception ex)
         {
             logger.LogWarning(ex,
--- a/src/gateway/BotNexus.Gateway/Agents/AgentExchangeService.cs
+++ b/src/gateway/BotNexus.Gateway/Agents/AgentExchangeService.cs
@@ -198,6 +198,13 @@ public sealed class AgentExchangeService : IAgentExchangeService
             await _sessionStore.SaveAsync(session, cancellationToken).ConfigureAwait(false);
             await ClearActiveSessionAsync(conversation, sessionId, CancellationToken.None).ConfigureAwait(false);
         }
+        catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
+        {
+            // #553: caller-initiated cancellation must NOT seal the session. See the matching
+            // comment in CrossWorldFederationController.ExecuteRelayAsync for full rationale —
+            // sealing here would poison the session for any caller retry.
+            throw;
+        }
         catch (Exception ex)
         {
             _logger.LogWarning(ex, "Agent conversation failed for session '{SessionId}'.", sessionId);
@@ -358,6 +365,12 @@ public sealed class AgentExchangeService : IAgentExchangeService
             session.Metadata["remoteSessionId"] = remoteSessionId;
             await _sessionStore.SaveAsync(session, cancellationToken).ConfigureAwait(false);
             await ClearActiveSessionAsync(conversation, sessionId, CancellationToken.None).ConfigureAwait(false);
+        }
+        catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
+        {
+            // #553: caller-initiated cancellation must NOT seal the session. See the matching
+            // comment in CrossWorldFederationController.ExecuteRelayAsync for full rationale.
+            throw;
         }
         catch (Exception ex)
         {
--- a/tests/architecture/BotNexus.Architecture.Tests/CancelNoSealArchitectureTests.cs
+++ b/tests/architecture/BotNexus.Architecture.Tests/CancelNoSealArchitectureTests.cs
@@ -1,0 +1,391 @@
+using System.Text.RegularExpressions;
+using Shouldly;
+
+namespace BotNexus.Architecture.Tests;
+
+/// <summary>
+/// Architecture fitness functions enforcing the <c>#553</c> contract: in
+/// <see cref="System.Threading.CancellationToken"/>-aware code paths that seal the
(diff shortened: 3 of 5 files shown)


sytone:
Filed deferred follow-up as #566 — duplicate user turn on retry after cancelled relay/exchange. Bounded but real; see PR body for trade-off rationale.

Issue #613, opened 2026-05-29
[Platform] Phase 9 — Conversation/Session/Agent model reconciliation


sytone:
# Phase 9 — Conversation/Session/Agent model reconciliation

Structural campaign to fix the conversation/session/agent layering so it's simpler for channels to interact with and removes duplicate-identity slop from the platform. Driven by the decisions in [#612 addendum](https://github.com/sytone/botnexus/issues/612#issuecomment-4569422852) on the 6 model gaps (G-1..G-6) + 6 design weaknesses (W-1..W-6) from the post-#579 architecture review.

**Sibling tracker**: #612 (mechanical cleanup waves W-1..W-X — distinct from this structural campaign).

## Core model after Phase 9

```
World
 └── Citizens (Users + Agents)
 └── Conversations (initiator + participants, KIND=UserAgent|AgentAgent|AgentSubAgent)
      └── Sessions (turn-history containers; derive participants from Conversation)
           └── SessionEntries (each stamped with Trigger=User|Cron|Soul|Heartbeat|AgentReply)
```

**Key shifts** (per addendum):
1. `Conversation` is the sole identity owner. `Session` becomes a strictly turn-history container.
2. Triggers (Cron/Soul/Heartbeat) are **proxy senders**, not session/conversation types — they are stamped on the inbound message / session entry, never on the Session or Conversation.
3. `Session.ConversationId` is non-nullable. No "legacy ungrouped" sessions — backfill to a per-agent Legacy Conversation.
4. `Conversation.WorldId` is required (parity with Agent).
5. Channels own the rendering choice (portal: list view; Telegram: prefix/switch/single-binding; iMessage: similar). Gateway publishes metadata only.

## Work items

| ID | Title | Driven by | Est size | Status |
|---|---|---|---|---|
| P9-A | `WorldId` on Conversation + migration | G-6 + W-6 | Small | — |
| P9-B | Legacy-Conversation backfill for orphan sessions; `Session.ConversationId` non-null | G-1 | Small/Med | — |
| P9-C | Auto-archive A↔A conversations on exchange end | W-3 | Small | — |
| P9-D | Invert CronJob ↔ Conversation link; friendly Conversation on first run; delete `cronconv:<agent>:<job>` composite-id; archive Conversation when CronJob deleted | G-5 | Med | — |
[...]

PR #548, opened 2026-05-26
feat(agents)!: route named↔named exchanges through IConversationStore (Phase 4 / F-3)

sytone:
## Summary

Phase 4 item 1a / F-3 of the BotNexus domain-model cleanup.

Closes the F-3 bug — named-agent ↔ named-agent exchanges never created a real `Conversation`, so they were invisible to `IConversationStore.ListByCitizenAsync`, the portal, and `ISessionStore.ListByConversationAsync` (F-7). `AgentExchangeService.ConverseAsync` now creates a real `Conversation` via `IConversationStore` and pins the child session to it eagerly (F-6 ordering) before any `PromptAsync` fires.

Both branches of `ConverseAsync` (local + cross-world) are migrated. The cross-world **receiver** side (`CrossWorldFederationController.RelayAsync`) is **not** touched in this PR — that's Phase 4 item 1b. Acknowledged tech debt: sender and receiver will hold different `ConversationId`s until 1b lands.

### Production code
- `Conversation.Kind` (new) — `ConversationKind { HumanAgent, AgentAgent, AgentSubAgent }`; defaults to `HumanAgent` so every existing row reads back with the same semantics.
- `AgentExchangeResult.ConversationId` (new, required, non-nullable) — forces every constructor path to supply the link.
- `AgentExchangeService.ConverseAsync` — rewritten to:
  1. Create `Conversation` via `IConversationStore.CreateAsync` (`Kind = AgentAgent`, `Initiator = CitizenId.Of(initiator)`, `AgentId = initiator`, `Title = "{init} ↔ {tgt}"`, `Purpose = objective`).
  2. Mint generic `SessionId.Create()` — no more `::agent-agent::` substring encoding.
  3. **Eager-pin** `session.Session.ConversationId = conv.ConversationId` + `SaveAsync` BEFORE `supervisor.GetOrCreateAsync` / `PromptAsync` (F-6 contract).
  4. Set `conversation.ActiveSessionId = sessionId` + save.
  5. Run the prompt loop, then seal + clear `ActiveSessionId`.
- `ConverseCrossWorldAsync` — same shape; propagates real `conversation.ConversationId.Value` through `OutboundMessage.Metadata["conversationId"]`.

### What did NOT change
[...]

PR #625, opened 2026-05-29
feat(gateway): auto-archive A↔A conversations on exchange end (P9-C)

### LINKED ISSUE #623: [Gateway] Phase 9 / P9-C — Auto-archive A↔A conversations on exchange end
**Umbrella**: #613 Phase 9. Driven by **W-3**: *""A-A conversations should have an end and then the conversation is done as that topic of conversation is over. So A-A conversations should be archived upon completion.""*

## Problem

Today, when `IAgentExchangeService.ConverseAsync` (or its cross-world sibling) finishes, the session is sealed and `Conversation.ActiveSessionId` is cleared — but `Conversation.Status` stays `Active`. The conversation lingers in the portal/list APIs forever, even though the agent-to-agent topic is over.

Per W-3, A↔A conversations are inherently bounded by an exchange. When the exchange ends (any reason except caller cancellation), the conversation is done. The agents created it for one purpose; that purpose is concluded.

## Scope

**In scope:**
- `AgentExchangeService.ConverseAsync` (local A↔A): archive conversation on every loop-exit path except caller cancellation
- `AgentExchangeService.ConverseCrossWorldAsync` (cross-world sender): same
- `CrossWorldFederationController.RelayAsync` (cross-world receiver): archive on `exchangeFinished` + `request.CloseAfterResponse` (new wire flag for sender-determined terminal turns) + error catch
- Wire-protocol extension: `CrossWorldRelayRequest.CloseAfterResponse` (bool, defaults false — backwards-compatible)

**Out of scope:**
- HumanAgent + AgentSubAgent conversations (operator-archived only, unchanged)
- Real-time portal ""archived"" notification on the new archive sites (existing `ConversationsController.Archive` does this via `NotifyConversationChangedBestEffortAsync`; bypassed for P9-C — documented limitation, defer to W-3 consolidation)
- Atomic `ArchiveIfActiveSessionAsync` store primitive (non-atomic check-then-archive accepted for P9-C since A↔A conversations are owned by AgentExchangeService and concurrent mutation is implausible; defer to W-3)

## Design (rubber-duck reviewed)

**New private helper** (duplicated in `AgentExchangeService` + `CrossWorldFederationController` for now; W-3 will consolidate):

```csharp
private async Task ArchiveOnExchangeEndAsync(Conversation conversation, SessionId expectedSessionId, CancellationToken cancellationToken)
{
    // P9-C: A↔A conversations auto-archive when the exchange ends. The store's ArchiveAsync
    // atomically nulls ActiveSessionId, so this subsumes the prior ClearActiveSessionAsync call.
    try
    {
        var latest = await _conversationStore.GetAsync(conversation.ConversationId, cancellationToken).ConfigureAwait(false);
        if (latest is null) return;
        if (latest.Status == ConversationStatus.Archived) return;                  // idempotent
        if (latest.ActiveSessionId != expectedSessionId) return;                   // someone else owns it now
        await _conversationStore.ArchiveAsync(conversation.ConversationId, cancellationToken).ConfigureAwait(false);
    }
    catch (Exception ex)
    {
        // Archive is a derived state; failing must not propagate as a ConverseAsync failure.
        _logger.LogWarning(ex, ""Failed to archive conversation '{ConversationId}' after exchange end."", conversation.ConversationId);
    }
}
```

**Wire-protocol change**:
```csharp
public sealed record CrossWorldRelayRequest
{
    // ... existing fields ...

    /// <summary>
    /// Sender-determined finality signal. When true, the sender has decided this is the
    /// final relay turn for the exchange (single-shot, or turn == MaxTurns-1) and the
    /// receiver should archive its conversation regardless of whether the target agent
    /// invoked finish_agent_exchange. Defaults to false — older sender builds without
    /// this flag will only trigger receiver archive via ExchangeFinished, preserving
    /// existing behavior. P9-C.
    /// </summary>
    public bool CloseAfterResponse { get; init; }
}
```

**Sender computes CloseAfterResponse** in `ConverseCrossWorldAsync` before each relay call:
`isFinalTurn = string.IsNullOrWhiteSpace(request.Objective) || turn == request.MaxTurns - 1;`

**Receiver checks** in `CrossWorldFederationController.RelayAsync`:
`if (exchangeFinished || request.CloseAfterResponse) ArchiveOnExchangeEndAsync(...) else ClearActiveSessionAsync(...)`

**Removed code**: `AgentExchangeService.ClearActiveSessionAsync` becomes dead (all 4 call sites flip to `ArchiveOnExchangeEndAsync`). Net production LOC: small positive (helper + wire field) minus the deleted `ClearActiveSessionAsync` minus the simpler call-site signatures.

## Acceptance

- [ ] `AgentExchangeService.ConverseAsync` (local A↔A): conversation status is `Archived` after every loop-exit path (exchangeFinished / singleShot / MaxTurns / error) except caller cancellation
- [ ] `AgentExchangeService.ConverseCrossWorldAsync` (sender side): same
- [ ] `CrossWorldFederationController.RelayAsync` (receiver side): conversation status is `Archived` when `exchangeFinished` OR `request.CloseAfterResponse` OR error; remains `Active` for non-final relays
- [ ] `ClearActiveSessionAsync` removed from `AgentExchangeService` (no callers); kept in `CrossWorldFederationController` (used for non-final relay turns)
- [ ] Caller-cancellation path leaves conversation `Active`
- [ ] Archive failures are logged at Warning and do NOT propagate as `ConverseAsync` failures
- [ ] Idempotent: re-calling the helper on an already-archived conversation is a no-op
- [ ] Pointer-mismatch safety: if a parallel actor has reassigned `ActiveSessionId` (or cleared it to null) between our `GetAsync` and intended archive, we skip — never clobber
- [ ] New `CrossWorldRelayRequest.CloseAfterResponse` defaults to `false` (backwards-compatible with older senders)
- [ ] Tests pin local + cross-world sender + cross-world receiver matrix including the receiver single-shot/MaxTurns cases, archive-failure isolation, idempotency, pointer-mismatch safety, and the no-archive-on-cancel invariant
- [ ] Architecture fitness function covering BOTH `AgentExchangeService` and `CrossWorldFederationController.RelayAsync` seal paths, with mandatory self-test (per stored convention)
- [ ] Reused `RemoteSessionId` after archive — receiver's existing sealed-session 409 guard already handles this; add a regression test
- [ ] Full `dotnet build BotNexus.slnx` + `dotnet test BotNexus.slnx` green; 0 warnings under `TreatWarningsAsErrors`

## Documented limitations (deferred to W-3 / future PR)

1. Direct `IConversationStore.ArchiveAsync` calls bypass the portal real-time ""archived"" notification emitted by `ConversationsController.Archive` via `NotifyConversationChangedBestEffortAsync`. Portal will see archived A↔A conversations on next list refresh / SignalR conversation-list sync, not as an immediate push. Acceptable for P9-C since A↔A conversations are typically short-lived; W-3 consolidation will route both paths through a shared lifecycle service that emits the notification.
2. Check-then-archive (`GetAsync` → `ArchiveAsync`) is non-atomic. A theoretical race exists where a parallel actor mutates `ActiveSessionId` after our check but before our archive. For A↔A conversations this is implausible (AgentExchangeService creates a fresh conversation per call; receiver-side reuse is serialized by sender). W-3 may add an atomic `ArchiveIfActiveSessionAsync` primitive when the same pattern is needed for HumanAgent conversations.

## Implementation order

1. Add `CloseAfterResponse` to `CrossWorldRelayRequest` (small, infrastructural)
2. Add `ArchiveOnExchangeEndAsync` helper to `AgentExchangeService`; flip 4 call sites; delete `ClearActiveSessionAsync`
3. Add `ArchiveOnExchangeEndAsync` helper to `CrossWorldFederationController`; branch the existing call site on `exchangeFinished || request.CloseAfterResponse`; add the archive in error catch
4. Wire the sender's `OutboundMessage.Metadata` `closeAfterResponse` → `CrossWorldRelayRequest.CloseAfterResponse` mapping in `CrossWorldChannelAdapter`
5. Tests (pin behavior + architecture fence)
6. Run full build + suite

Closes after merge.

sytone:
Implements **W-3** ("A-A conversations should have an end and then the conversation is done") by auto-archiving named↔named agent-exchange conversations when the exchange terminates. Closes #623; refs umbrella #613.

## Problem

Pre-P9-C, every named↔named exchange minted a real `Conversation` (Phase 4 / F-3, PRs #548 sender + #549 receiver) but only ever sealed the underlying session and cleared `ActiveSessionId`. The conversation itself stayed `Active` forever, accumulating dead entries in portal/list APIs.

## What this PR does

Four terminal seal paths now archive the conversation in addition to sealing the session:

| Path | File:line |
|------|-----------|
| Local A↔A success seal | `AgentExchangeService.ConverseAsync` (success branch) |
| Local A↔A error catch | `AgentExchangeService.ConverseAsync` (catch-all) |
| Cross-world sender success seal | `AgentExchangeService.ConverseCrossWorldAsync` (success branch) |
| Cross-world sender error catch | `AgentExchangeService.ConverseCrossWorldAsync` (catch-all) |
| Cross-world receiver — finish-tool branch | `CrossWorldFederationController.ExecuteRelayAsync` (split success branch) |
| Cross-world receiver — `CloseAfterResponse` branch | same (split success branch) |
| Cross-world receiver — error catch | same (catch-all) |

**Non-final cross-world relays** still take the existing `ClearActiveSessionAsync` path (no archive) — the exchange continues.

### Wire protocol addition

`CrossWorldRelayRequest.CloseAfterResponse : bool` lets the sender signal "this is the final turn" so the receiver archives even when the target agent never invokes `finish_agent_exchange` (single-shot + max-turns cases). Lifted from `OutboundMessage.Metadata["closeAfterResponse"]` by `CrossWorldChannelAdapter.TryGetMetadataBool`. Defaults to `false` so older senders preserve pre-P9-C behaviour (receiver archives only on `ExchangeFinished`).

## Test coverage matrix

| Layer | File | Count |
|---|---|---|
| Behaviour pins — local sender | `tests/gateway/BotNexus.Gateway.Tests/Agents/AgentExchangeArchiveTests.cs` | 8 |
| Behaviour pins — cross-world receiver | `tests/gateway/BotNexus.Gateway.Tests/CrossWorldFederationControllerTests.cs` (inline) | 6 |
| Behaviour pins — adapter metadata lift | `tests/gateway/BotNexus.Gateway.Tests/Channels/CrossWorldChannelAdapterTests.cs` | 9 |
| Architecture fence | `tests/architecture/BotNexus.Architecture.Tests/AgentExchangeArchiveArchitectureTests.cs` | 8 |
| **Total new tests** | | **31** |

All pass. Full Gateway.Tests: 1990/1 skip. Full Architecture.Tests: 102/102. Build: 0 warnings / 0 errors (TreatWarningsAsErrors).

## Multi-model critique sweep applied

Three independent agents (security-review/gpt-5.5, plan-vs-impl/claude-opus-4.7, bug-hunt/gpt-5.3-codex) reviewed before merge. Fold-ins (commit ``772ccb17``):

- **HIGH** (bug-hunt + security defensive-hardening) — `TryGetMetadataBool` now handles `JsonElement(True/False/String)` shapes that arise when `OutboundMessage.Metadata` round-trips through Sqlite/File stores. Without this, a future change that persists OutboundMessages would silently regress to false → receiver wouldn't archive on the `CloseAfterResponse` path.
- **SHOULD-CONSIDER** (plan-vs-impl F1) — added 2 sender-side cross-world behaviour pins (success + error catch) that were previously fence-only.
- **SHOULD-CONSIDER** (plan-vs-impl F2) — added receiver-side failure-isolation pin via `ThrowingArchiveConversationStore`.

## Accepted limitations (deferred to W-3)

- **Helper duplication** — `ArchiveOnExchangeEndAsync` exists in both `AgentExchangeService.cs` and `CrossWorldFederationController.cs`. Consolidation deferred; intentional for P9-C scope. Helper bodies are kept literally identical to make drift easy to spot.
- **`IConversationStore.ArchiveIfActiveSessionAsync` atomic primitive** — current implementation does `GetAsync` → strict pointer guard → `ArchiveAsync`. There's a documented (and tested) TOCTOU window between the guard and the archive write. In production every caller mints its own `ConversationId`, so the window is non-exploitable today; an atomic primitive would close it for future shared-conversation flows.
- **SignalR `NotifyConversationChangedBestEffortAsync("archived")`** — the new archive sites do not emit a SignalR fan-out event. Portal currently picks up the state change via the next list-refresh. A push notification can land alongside W-3's shared lifecycle service.
- **Receiver state-machine quirk** — when `CloseAfterResponse=true && exchangeFinished=false`, the conversation archives but the session is **not** sealed. A follow-up relay against the same `RemoteSessionId` would resurrect `ActiveSessionId` on an Archived conversation. Security review flagged this as portal/UX inconsistency only (no auth bypass, no cross-tenant leak). To be filed as a separate cleanup issue.

## Commits

- `ff59a721` feat(gateway): auto-archive A↔A conversations on exchange end (P9-C)
- `4e84519a` test(gateway): pin A↔A auto-archive across local + cross-world + receiver paths (P9-C)
- `8b7bfde2` test(architecture): fence A↔A archive helper on exchange seal paths (P9-C)
- `772ccb17` fix(gateway): harden P9-C metadata lift + close coverage gaps from critique sweep

Closes #623
Refs #613

### DIFF
--- a/src/domain/BotNexus.Domain/Gateway/Models/CrossWorldRelayModels.cs
+++ b/src/domain/BotNexus.Domain/Gateway/Models/CrossWorldRelayModels.cs
@@ -34,6 +34,17 @@ public sealed record CrossWorldRelayRequest
     /// Gets or sets the remote session id.
     /// </summary>
     public string? RemoteSessionId { get; init; }
+
+    /// <summary>
+    /// Sender-determined finality signal (Phase 9 / P9-C). When <c>true</c>, the sender has
+    /// decided this is the final relay turn for the exchange — either because the request
+    /// was single-shot (no <c>Objective</c>) or because <c>turn == MaxTurns - 1</c>. The
+    /// receiver MUST archive its local conversation when the relay completes regardless of
+    /// whether the target agent invoked <c>finish_agent_exchange</c>. Defaults to <c>false</c>
+    /// so older senders without the signal preserve existing behaviour (receiver archives
+    /// only when <c>ExchangeFinished</c> is true).
+    /// </summary>
+    public bool CloseAfterResponse { get; init; }
 }
 
 /// <summary>
--- a/src/gateway/BotNexus.Gateway.Api/Controllers/CrossWorldFederationController.cs
+++ b/src/gateway/BotNexus.Gateway.Api/Controllers/CrossWorldFederationController.cs
@@ -218,12 +218,19 @@ public sealed class CrossWorldFederationController(
             }
             await sessionStore.SaveAsync(session, cancellationToken).ConfigureAwait(false);
 
-            // Leave the session Active so the sender can continue the exchange by supplying
-            // RemoteSessionId on the next turn; clear ActiveSessionId so portal stops rendering
-            // the conversation as "in flight" while the sender pauses between turns. When the
-            // exchange has finished, the session is already Sealed above and ResolveSessionAsync
-            // will reject any further reuse.
-            await ClearActiveSessionAsync(conversation, sessionId, CancellationToken.None).ConfigureAwait(false);
+            // P9-C: when the exchange is ending — either because the target invoked
+            // finish_agent_exchange (exchangeFinished) OR the sender signalled this is the
+            // final turn via CloseAfterResponse (single-shot / max-turns) — archive the
+            // receiver-side conversation. Otherwise just clear ActiveSessionId so the portal
+            // stops rendering it as "in flight" while the sender pauses between turns.
+            if (exchangeFinished || request.CloseAfterResponse)
+            {
+                await ArchiveOnExchangeEndAsync(conversation, sessionId, CancellationToken.None).ConfigureAwait(false);
+            }
+            else
+            {
+                await ClearActiveSessionAsync(conversation, sessionId, CancellationToken.None).ConfigureAwait(false);
+            }
 
             return Ok(new CrossWorldRelayResponse
             {
@@ -262,7 +269,10 @@ public sealed class CrossWorldFederationController(
             session.Status = GatewaySessionStatus.Sealed;
             session.Metadata["error"] = ex.Message;
             await sessionStore.SaveAsync(session, CancellationToken.None).ConfigureAwait(false);
-            await ClearActiveSessionAsync(conversation, sessionId, CancellationToken.None).ConfigureAwait(false);
+            // P9-C: failed exchanges are terminal — archive rather than just clearing the
+            // pointer. The session is already sealed; the conversation should not linger as
+            // Active in portal/list APIs after a failure.
+            await ArchiveOnExchangeEndAsync(conversation, sessionId, CancellationToken.None).ConfigureAwait(false);
             throw;
         }
     }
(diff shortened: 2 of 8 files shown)


sytone:
Filed separate follow-up issue #626 for the receiver state-machine quirk (CloseAfterResponse=true && exchangeFinished=false → conversation Archived but session Active). Out of scope for this PR; tracked for cleanup.

Issue #676, opened 2026-05-31
[Channels] Phase 10 — channels own their conversation binding strategy (W-5 umbrella)


sytone:
## Phase 10 — Channels own their conversation binding strategy (W-5)

Phase 9 (issues #614…#674, 16 sub-PRs) made `Conversation` the durable owner of agent identity, participants, and bindings, and reduced `Session` to a pure ""active window of work"" inside a Conversation. The substrate is now clean.

But on the **channel side**, multiple legacy patterns have accumulated that bypass or duplicate the gateway routing primitives. Per the W-5 directive:

> Each channel can choose how conversations are bound and rendered. That is a channel responsibility — the gateway should provide enough information for the channel to make this effective. The gateway should not care HOW a channel chooses to bind.

This issue is the umbrella for cleaning that up.

## What's wrong today

1. **`IChannelAdapter.SendStreamDeltaAsync(string conversationId, …)` is a lie.** The caller (`GatewayHost.cs:590`) passes `message.ChannelAddress.Value`. SignalR interprets it as a `SessionId`. Telegram parses it back as a `ChannelAddress` and decodes to `chatId`. TUI ignores it. It's actually a per-channel opaque routing token, not a conversation id.

2. **`GatewayHub.cs` is a 770-LOC binding monolith** with 13 injected services. It hand-implements `ResolveOrCreateSessionAsync` (lines 701-777) re-doing what `IConversationDispatcher` already does. It directly mutates session fields and calls `IConversationStore.AddParticipantsAsync`. Two `[Obsolete]` hub methods (`JoinSession`/`LeaveSession`) violate the codebase's no-`[Obsolete]` rule.

3. **`TuiChannelAdapter` hardcodes `RequestedSessionId = SessionId.From(""tui-console"")`** (lines 200-201, 219-220). Pre-P9 escape hatch — bypasses the natural conversation→session resolution.

4. **Three implicit `ChannelAddress` conventions** — Portal uses `agentId` as address, Telegram uses `chatId[:threadId]`, TUI uses `""console""` — none declared, all baked into adapter behaviour.

5. **`InternalChannelAdapter.ResolveTargetAdapterForConversationAsync`** is a parallel resolution path duplicating the binding system.

6. **ServiceBus has its own `_pendingReplies`/`_pendingQueue` reply-context complexity** that no other channel models — a fifth implicit binding pattern.

## Plan: 6 sub-PRs

| # | Scope | LOC | Risk |
|---|---|---|---|
[...]

PR #679, opened 2026-05-31
feat(channels)!: replace string conversationId with typed ChannelStreamTarget on stream adapter methods

### LINKED ISSUE #677: [Channels] Phase 10 PR1 — typed ChannelStreamTarget replacing string conversationId on stream methods
## Parent: #676

Phase 10 PR1 — Replace the misleading `string conversationId` stream parameter with a typed `ChannelStreamTarget` record.

## Problem

`IChannelAdapter.SendStreamDeltaAsync(string conversationId, string delta, ct)` and `IStreamEventChannelAdapter.SendStreamEventAsync(string conversationId, AgentStreamEvent, ct)` both take a parameter named `conversationId` that is NOT a conversation id.

Today the caller `GatewayHost.cs:590` passes `message.ChannelAddress.Value`. Each channel interprets the value differently:

| Channel | Interpretation |
|---|---|
| SignalR | `SessionId.From(value)` — SignalR group key |
| Telegram | `ChannelAddress.From(value)` then `TelegramChannelAddress.TryDecode` to `(chatId, threadId)` |
| TUI | Ignored (used only as a label in stream-event display) |
| Internal | Passed through to the resolved target adapter |
| Observer fan-out (`GatewayHost.cs:611`) | `binding.ChannelAddress.Value` — different from the active session's channel address |

This is fragile and post-P9-incoherent. The parameter is actually a per-channel opaque routing token.

## Solution

Introduce a new record in `BotNexus.Domain.Gateway.Models`:

```csharp
public sealed record ChannelStreamTarget(
    SessionId SessionId,
    ChannelAddress ChannelAddress,
    BindingId? BindingId = null);
```

Rename the two adapter methods:

```csharp
// IChannelAdapter
Task SendStreamDeltaAsync(ChannelStreamTarget target, string delta, CancellationToken ct = default);

// IStreamEventChannelAdapter
Task SendStreamEventAsync(ChannelStreamTarget target, AgentStreamEvent streamEvent, CancellationToken ct = default);
```

Update each adapter to consume the field that matches its semantics:

| Adapter | Uses |
|---|---|
| SignalR | `target.SessionId` (or `streamEvent.SessionId` if set, current behaviour preserved) |
| Telegram | `target.ChannelAddress` (decoded via `TelegramChannelAddress.TryDecode`) |
| TUI | `target.SessionId.Value` for display label only |
| Internal | Pass full `target` to the resolved target adapter |
| Virtual (scenario harness) | Record the target for assertions |

Update `GatewayHost` callers to construct the target with the live `SessionId` + `message.ChannelAddress` + originating `BindingId`. Update observer fan-out path to construct per-binding targets.

## Out of scope

- Hub slim-down (#TBD, will follow as PR4 of the Phase 10 arc).
- TUI hardcoded sessionId removal (will follow as PR2 of the Phase 10 arc).
- Binding policy declarations (will follow as PR6 of the Phase 10 arc).

## Tests

- Update each adapter's unit tests to use the new signature.
- Add tests verifying:
  - SignalR routes by `target.SessionId`.
  - Telegram routes by `target.ChannelAddress` (current chatId decode path still works).
  - Internal correctly passes the full `ChannelStreamTarget` to the resolved target.
  - `GatewayHost` constructs targets correctly for both active-session and observer fan-out paths.
- Run architecture tests (must still pass — `VirtualChannelAdapter` updated to match).

## Acceptance

- Build green, zero warnings (warnings-as-errors active).
- Full test suite green.
- No remaining usages of the string-typed `SendStreamDeltaAsync`/`SendStreamEventAsync` signatures.
- `InternalsVisibleTo` / scenario harness compiles against the new signatures.

sytone:
**Part of W-5 / Phase 10 channel cleanup arc** — umbrella: #676. Tracking issue: #677.

## What

Replaces the loosely-typed `string conversationId` parameter on `IChannelAdapter.SendStreamDeltaAsync` and `IStreamEventChannelAdapter.SendStreamEventAsync` with a typed `ChannelStreamTarget(SessionId, ChannelAddress, BindingId?)` record.

This is **PR1 of 6** in the W-5 channel-binding cleanup. It surgically removes the per-adapter ambiguity on the streaming surface so subsequent PRs can simplify routing, dispatching, and observer fan-out without that ambiguity hiding bugs.

## Why

The old `string conversationId` meant different things to different adapters:

| Adapter | What it treated the string as |
|---|---|
| SignalR | `SessionId` (SignalR group key) |
| Telegram | `ChannelAddress` (with composite `chat-id/topic:NN` suffix) |
| TUI | Ignored — used only as a display label |
| InternalChannelAdapter | Tried to parse as a `SessionId`, then fell through to a SignalR fallback when parsing yielded "no such session" |

The cleanup directive says: *"too much complexity in the interactions between the code layers and channels."* This is one of the slop fragments. Removing it now keeps PR2–PR6 honest.

## Behaviour fix (NOT scope creep — please read)

`InternalChannelAdapter.ResolveTargetAdapterForSessionAsync(SessionId, ct)` now uses the typed `SessionId` from `ChannelStreamTarget` to look up the parent session. Previously, the parameter was a string parsed as `SessionId`, but the gateway was actually passing the *channel address* (e.g. `chat-200/topic:99`). The lookup silently failed for any non-SignalR parent session and fell through to the SignalR fallback adapter — so a sub-agent wake-up event for a Telegram-rooted parent session was getting delivered to SignalR instead of Telegram.

**The new behaviour is pinned by a dedicated test** (`InternalChannelAdapterTests.SendStreamEventAsync_ResolvesByTypedSessionId_AndRoutesToParentsNonSignalRChannel`) so a future regression would fail CI.

### New
- `src/domain/BotNexus.Domain/Gateway/Models/ChannelStreamTarget.cs` — typed record, documents how each adapter consumes each field.
- `tests/architecture/.../ChannelStreamTargetArchitectureTests.cs` — 4 fence tests that prevent re-introduction of a parallel string overload.
- `tests/gateway/.../_TestUtilities/StreamTargets.cs` — internal test helper for the common "I don't care about distinguishing SessionId vs ChannelAddress in this unit test" pattern.

### Adapters updated
- `SignalRChannelAdapter` — uses `target.SessionId`.
- `TelegramChannelAdapter` — uses `target.ChannelAddress` (preserves composite topic suffix).
- `TuiChannelAdapter` — uses `target.SessionId` as display label.
- `InternalChannelAdapter` — resolves parent's true adapter via `target.SessionId` (behaviour fix above).
- `ChannelAdapterBase` — virtual signature updated.
- `VirtualChannelAdapter` (harness) — keys captured deltas/events by `target.ChannelAddress.Value`; also exposes a new `StreamTargets` snapshot for tests that want to assert the full typed payload.

### Production callers
- `GatewayHost` — both the primary streaming path and the SignalR observer fan-out now construct `ChannelStreamTarget` with the resolved `SessionId`, the originating/observer `ChannelAddress`, and the originating/observer `BindingId`. The observer tuple was widened from `(adapter, string sessionAddress)` to `(adapter, ChannelAddress, BindingId)` so the per-observer typed target can be built inside the foreach. The `SendAskUserToBindingAsync` helper is similarly updated.

### Docs
- `docs/architecture/system-flows.md`, `src/gateway/README.md`, `src/gateway/BotNexus.Gateway.Channels/README.md` — example signatures + per-field consumption table.

## Verification

```
Build:            0 errors / 0 warnings   (warnings-as-errors active)
Impacted tests:   17 projects, ALL GREEN
Gateway tests:    2005 pass / 1 skip / 0 fail   (+1 new behaviour-fix test)
Architecture:     161 pass / 0 fail              (+4 new fence tests)
Scenarios:        29 pass / 0 fail
```

Run locally:
```pwsh
scripts/repo/build.ps1
scripts/repo/test-impacted.ps1
```

## Out of scope (the remaining 5 PRs in #676)

- PR2: TUI cleanup — remove `RequestedSessionId = "tui-console"` hardcode
- PR3: Inbound message orchestrator — collapse Router → Dispatcher into a single typed entry point
- PR4: Channel-side binding ownership — channels register their preferred binding strategy declaratively
- PR5: GatewayHub slim-down — 770 → ~200 LOC by delegating to the orchestrator
- PR6: Tidy implicit ChannelAddress conventions; introduce `ChannelBindingPolicy` record

Each ships once PR1 merges so reviewers don't see overlapping diffs.

Closes #677.

### DIFF
--- a/docs/architecture/system-flows.md
+++ b/docs/architecture/system-flows.md
@@ -209,7 +209,8 @@ AgentHandle → ChannelAdapter.SendAsync(OutboundMessage) → SignalR Group Broa
 
 1. **Agent emits `AgentEvent`** (e.g., `ContentDelta`, `ToolStart`, `MessageEnd`)
 2. **Handle wraps as `AgentStreamEvent`** with sessionId
-3. **Channel adapter receives event**: `IChannelAdapter.SendStreamDeltaAsync(conversationId, delta, ct)`
+3. **Channel adapter receives event**: `IChannelAdapter.SendStreamDeltaAsync(ChannelStreamTarget target, string delta, CancellationToken ct)`
+   The `ChannelStreamTarget` carries the typed `SessionId`, `ChannelAddress`, and optional `BindingId` the adapter needs to route delivery — each adapter consumes whichever field matches its native routing model (SignalR groups by `SessionId`, Telegram looks up by `ChannelAddress`, etc.).
 4. **Adapter broadcasts to session group**: `Clients.Group("session:{sessionId}").SendAsync(method, data)`
 5. **All clients in group receive event** in real-time
 
--- a/src/domain/BotNexus.Domain/Gateway/Models/ChannelStreamTarget.cs
+++ b/src/domain/BotNexus.Domain/Gateway/Models/ChannelStreamTarget.cs
@@ -1,0 +1,63 @@
+using BotNexus.Domain.Primitives;
+
+namespace BotNexus.Gateway.Abstractions.Models;
+
+/// <summary>
+/// Typed routing token passed to <c>IChannelAdapter.SendStreamDeltaAsync</c> and
+/// <c>IStreamEventChannelAdapter.SendStreamEventAsync</c>. Each channel adapter
+/// consumes the field that matches its routing semantics:
+/// </summary>
+/// <remarks>
+/// <para>
+/// Replaces the historical <c>string conversationId</c> parameter that grew out of
+/// pre-Phase 9 channel modelling. That string was an opaque per-channel token —
+/// SignalR interpreted it as a <see cref="SessionId"/>, Telegram parsed it back to
+/// a <see cref="ChannelAddress"/>, the internal fan-out path passed each binding's
+/// channel address. The misleading parameter name and overloaded semantics were a
+/// recurring source of routing bugs; this record names the three facets explicitly
+/// so each adapter picks the one it needs.
+/// </para>
+/// <para>
+/// Field consumption per built-in adapter:
+/// <list type="bullet">
+///   <item><b>SignalR</b> — uses <see cref="SessionId"/> to address the SignalR
+///   group. Stream events that carry their own <c>SessionId</c> still take
+///   precedence so cross-session enrichment continues to work.</item>
+///   <item><b>Telegram</b> — uses <see cref="ChannelAddress"/> to recover the
+///   native <c>chatId</c>/<c>messageThreadId</c> via
+///   <c>TelegramChannelAddress.TryDecode</c>.</item>
+///   <item><b>TUI</b> — uses <see cref="SessionId"/> only as a display label.</item>
+///   <item><b>Internal</b> — uses <see cref="SessionId"/> to resolve the parent
+///   session's original channel adapter and forwards the same target verbatim.</item>
+///   <item><b>Observer fan-out</b> — caller constructs one target per observer
+///   binding so the originating <see cref="BindingId"/> can be excluded from echo
+///   and per-binding addresses are preserved.</item>
+/// </list>
+/// </para>
+/// <para>
+/// The record mirrors the shape of <see cref="OutboundMessage"/>'s addressing
+/// fields so callers can lift the same values across delivery and streaming
+/// paths without re-deriving anything.
+/// </para>
+/// </remarks>
+/// <param name="SessionId">
+/// The session the stream events belong to. Always set — channels that route by
+/// session use this directly; channels that route by channel address still need
+/// the session id for logging and event enrichment.
+/// </param>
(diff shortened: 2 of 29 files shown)


sytone:
**Heads-up for reviewers** — while triaging this PR a related finding came out: `SignalRChannelAdapter` keys its delivery group by `SessionId`, which means streaming silently drops across a compaction boundary (UI stays subscribed to the old session's group). That's intentionally **out of scope for this PR** (which is a typed rename + one flagged Internal fix) and tracked separately as **#682**, which will extend `ChannelStreamTarget` with a `ConversationId` field and switch SignalR's group key.

Filing it up-front so this PR's deliberately-conservative `SessionId`-only target is read as ""stage 1 of 2"" rather than ""we forgot."" Umbrella #676 has been updated with the renumbered plan.

PR #821, opened 2026-06-04
feat(gateway): fire SignalR push on ConversationTool metadata mutations

sytone:
Closes #697

## Summary

Wire IConversationChangeNotifier into ConversationTool so that conversation metadata mutations (title, purpose, instructions) and archive events are immediately pushed to connected portal clients via SignalR. Previously the sidebar only refreshed as a side-effect of message activity.

## Root Cause

ConversationTool mutated conversations via IConversationStore but never called IConversationChangeNotifier. The notifier was already wired into ConversationsController (REST path) and InProcessIsolationStrategy injected the store and dispatcher but not the notifier.

## Changes

- ConversationTool.cs:
  - Added optional IConversationChangeNotifier changeNotifier constructor parameter
  - Added NotifyBestEffortAsync private helper (swallows exceptions - SignalR failure must not fail tool calls)
  - SetTitleAsync: fires updated notification after save
  - SetPurposeAsync: fires updated notification after save
  - SetAsync: fires updated notification after save
  - ArchiveAsync: fires archived notification after archive
- InProcessIsolationStrategy.cs:
  - Resolves IConversationChangeNotifier from DI and passes to ConversationTool constructor
- ConversationToolTests.cs: 6 new tests
  - SetTitle_FiresUpdatedNotification
  - SetPurpose_FiresUpdatedNotification
  - Set_FiresUpdatedNotification
  - Archive_FiresArchivedNotification
  - SetTitle_WithoutNotifier_DoesNotThrow (no changeNotifier = no-op)
  - NotifierFailure_DoesNotPropagate_SetTitle (exception swallowed, mutation persists)

## Test Results

- ConversationToolTests: 28/28 pass (includes all tests from stacked #820)
- Architecture tests: 167/167 pass

## Merge Notes

This PR stacks on #820 (feat(gateway): add archive action to ConversationTool).
Merge order: #820 first, then this PR. Both touch ConversationTool.cs but this branch already includes #820's changes.
Safe to merge independently of #816 and #817 (no file overlap).

PR #866, opened 2026-06-05
fix(cron): fire SignalR notify when CronTrigger reactivates an archived pinned conversation

### LINKED ISSUE #864: [Gateway] Auto-unarchive conversation when a new message arrives on an archived conversation
## Summary

When a message arrives targeting an archived conversation (e.g. a cron job fires on a conversation that was previously archived, or a user opens an archived conversation and sends a message), the conversation should automatically be restored to `Active` status rather than silently failing or routing incorrectly.

## Motivation

Cron jobs use stable conversation IDs (`cronconv:agentId:jobId`). If such a conversation gets archived (e.g. during a cleanup pass), the next cron run will attempt to deliver a message to an archived conversation. The expected behaviour is transparent rehydration — the conversation becomes active again and the message is delivered normally.

Same applies to user-initiated messages: if a user navigates to an archived conversation in the portal and sends a message, it should rehydrate rather than fail.

## Acceptance Criteria

- [ ] When `IConversationStore.GetOrCreateAsync` (or equivalent dispatch path) resolves a conversation with `Status = Archived`, it automatically sets the status back to `Active` before proceeding
- [ ] A SignalR push notifies the portal that the conversation is now active (so the UI reflects the change without a page reload)
- [ ] The rehydration is idempotent — sending multiple messages in quick succession does not create duplicate conversations
- [ ] Existing archived conversations that are NOT messaged remain archived
- [ ] Unit test: assert that dispatching a message to an archived conversation results in `Status = Active` after dispatch

## Notes

- The trigger path to investigate: `InboundMessageOrchestrator` → session/conversation resolution → `IConversationStore`
- `cronconv:` conversations are the primary motivating case but the fix should be generic
- Related: cron jobs that are deleted should have their conversations archived (separate cleanup concern)

sytone:
Closes #864

## Problem

When a cron job fires against a pinned conversation that has been archived, `CronTrigger.ResolveOrCreateConversationAsync` correctly reactivated the status and saved the conversation — but did not fire a `IConversationChangeNotifier` notification. The portal sidebar would therefore show the conversation as archived until the next page load.

## Changes

- `CronTrigger.cs`: added optional `IConversationChangeNotifier?` parameter (backward-compat default = `null`). After reactivation, calls `NotifyConversationChangedAsync("updated", ...)` best-effort (exceptions are caught and logged as Warning, never propagated to the cron turn).
- `CronTriggerTests.cs`: 2 new regression tests
  - `CreateSessionAsync_PinnedArchivedConversation_FiresReactivationNotify` — verifies notify fires exactly once with correct parameters
  - `CreateSessionAsync_PinnedArchivedConversation_NoNotifier_DoesNotThrow` — verifies backward compat when no notifier is wired

## Test Results

`test-impacted.ps1`: 167 Architecture + 2077 Gateway.Tests + 125 Conversations.Tests + 29 Scenarios = all pass.

## Merge Notes

Independent of all open PRs — only touches `CronTrigger.cs` and `CronTriggerTests.cs`. Safe to merge in any order.

### DIFF
--- a/src/gateway/BotNexus.Gateway.Api/Triggers/CronTrigger.cs
+++ b/src/gateway/BotNexus.Gateway.Api/Triggers/CronTrigger.cs
@@ -27,7 +27,8 @@ public sealed class CronTrigger(
     IAgentSupervisor supervisor,
     IConversationStore conversations,
     ISessionStore sessions,
-    ILogger<CronTrigger> logger) : IInternalTrigger
+    ILogger<CronTrigger> logger,
+    IConversationChangeNotifier? changeNotifier = null) : IInternalTrigger
 {
     /// <summary>
     /// Gets the trigger type identifier.
@@ -139,6 +140,25 @@ public sealed class CronTrigger(
                     pinned.Status = ConversationStatus.Active;
                     pinned.UpdatedAt = DateTimeOffset.UtcNow;
                     await conversations.SaveAsync(pinned, ct).ConfigureAwait(false);
+
+                    // Notify the portal so the conversation reappears in the sidebar without
+                    // requiring a page reload. Best-effort: failure here must not block the
+                    // cron turn. (#864)
+                    if (changeNotifier is not null)
+                    {
+                        try
+                        {
+                            await changeNotifier.NotifyConversationChangedAsync(
+                                "updated", agentId.Value, pinned.ConversationId.Value, ct)
+                                .ConfigureAwait(false);
+                        }
+                        catch (Exception ex)
+                        {
+                            logger.LogWarning(ex,
+                                "CronTrigger: non-fatal failure sending reactivation notification for conversation {ConversationId}",
+                                pinned.ConversationId);
+                        }
+                    }
                 }
                 return pinned;
             }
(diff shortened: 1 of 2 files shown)


sytone:
## CI Monitor - 2026-06-05 11:23 PT

**Status:** All checks PASSING

behindBy: 0 | mergeable: MERGEABLE | isDraft: false

No action required - ready for Jon to merge.

sytone:
**CI:** impacted-tests ✅ pass (3m17s) | All checks ✅ PASSING
**Conflicts:** None (behindBy=0, MERGEABLE/CLEAN)
**Title:** `fix(cron): fire SignalR notify when CronTrigger reactivates an archived pinned conversation` — conventional commits ✅
**Coverage:** 2 tests — `CreateSessionAsync_PinnedArchivedConversation_FiresReactivationNotify` (happy path: notify fires once with correct params) and `CreateSessionAsync_PinnedArchivedConversation_NoNotifier_DoesNotThrow` (backward compat: no notifier, no throw). Both paths covered.
**Spec alignment:** Closes #864. CronTrigger now fires SignalR push on reactivation, consistent with ConversationTool notifier pattern (PR #821). Best-effort exception handling is correct — cron turn should never fail due to a notify error.
**Assessment:** LGTM — clean, targeted, well-tested.

Issue #1273, opened 2026-06-12
fix(config): titling field changed from string to object without migration, crashes on startup


sytone:
## Bug

The gateway crashes on startup with:
```
Invalid JSON in config.json. The JSON value could not be converted to System.String.
Path: $.gateway.auxiliary.titling | LineNumber: 88
```

### Root Cause

`Titling` was changed from a plain `string` (model name) to `TitlingConfig` object (with `Model` + `TimeoutSeconds`). Existing configs that have `"titling": "gpt-4o-mini"` now fail deserialization.

### Fix

Add migration in `PlatformConfigLoader.MigrateLegacyGatewaySettings()`:
- If `titling` is a string value → convert to `{ "model": "<value>" }`

### Impact

- Gateway cannot start with pre-existing config
- Users who configured titling model now crash on update
- **This likely caused recent 500 issues after code update**

### Labels
bug, priority:critical

sytone:
PR opened: https://github.com/sytone/botnexus/pull/1278

PR #1277, opened 2026-06-12
fix(gateway): remove IConfiguration.Bind() to prevent Docker runtime key collision

### LINKED ISSUE #1271: fix(gateway): PlatformConfig binding collides with DOTNET_VERSION env var in Docker containers
## Bug

The gateway **cannot start in a Docker container** on current main (`dfbaeb80`).

### Root Cause

`GatewayServiceCollectionExtensions.LoadConfigForRegistration()` (line 364) calls:
```csharp
configuration.Bind(config);
```

This binds the **entire** `IConfiguration` root into `PlatformConfig`. In a container, the .NET runtime injects `runtimeOptions:framework:version = "10.0.9"` into the config hierarchy. Since `PlatformConfig.Version` is `int`, the binder throws:

```
System.InvalidOperationException: Failed to convert configuration value '10.0.9' at 'Version' to type 'System.Int32'.
```

### Fix Options

1. **Best**: Bind to a specific section instead of root — `configuration.GetSection("botnexus").Bind(config)` or similar
2. Change `PlatformConfig.Version` to `string` (breaking)
3. Add `[BindNever]` attribute to `Version` property
4. Use `BinderOptions { ErrorOnUnknownConfiguration = false }` + try/catch (masks other issues)

### Impact

- **CRITICAL** — gateway cannot start in Docker at all
- docker-compose.yml is non-functional on main
- Blocks containerized deployment and CI validation

### Repro

```bash
docker build -t botnexus:test .
docker run -p 5000:5000 -v /tmp/cfg:/app/config botnexus:test
# Immediate crash
```

### Labels
bug, priority:critical

sytone:
Closes #1271

## Changes
- Replaced \configuration.Bind(config)\ in \LoadConfigForRegistration\ with direct file-based loading via \PlatformConfigLoader.Load()\
- The IConfiguration Bind approach was binding runtime-injected keys (e.g. \Version = 10.0.9\ from .NET framework config) to \PlatformConfig.Version\ (int), causing conversion crash in Docker
- File-based loading correctly handles migration and validation without exposure to ambient IConfiguration keys
- 2 new tests confirming load works with conflicting keys and missing files

## Merge Notes
Independent of all other open PRs. Safe to merge in any order. Should merge before #1273 if both ready simultaneously (both fix startup crashes).

### DIFF
--- a/src/gateway/BotNexus.Gateway/Extensions/GatewayServiceCollectionExtensions.cs
+++ b/src/gateway/BotNexus.Gateway/Extensions/GatewayServiceCollectionExtensions.cs
@@ -357,19 +357,11 @@ public static class GatewayServiceCollectionExtensions
 
     private static PlatformConfig LoadConfigForRegistration(IConfiguration? configuration, string resolvedConfigPath, IFileSystem fileSystem)
     {
-        if (configuration is null)
-            return PlatformConfigLoader.Load(resolvedConfigPath, fileSystem: fileSystem);
-
-        var config = new PlatformConfig();
-        configuration.Bind(config);
-        var rawJson = TryReadConfigFile(resolvedConfigPath, fileSystem);
-        if (!string.IsNullOrWhiteSpace(rawJson))
-        {
-            PlatformConfigLoader.MigrateLegacyGatewaySettings(config, rawJson);
-            PlatformConfigLoader.ExtractAgentDefaults(config, rawJson);
-        }
-
-        return config;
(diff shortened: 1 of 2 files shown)


sytone:
Closing — the fix bypasses IConfiguration entirely which is wrong. Reopening with a proper fix that uses sectioned binding per the standard IConfiguration/IOptions pattern. Root cause is DOTNET_VERSION env var getting prefix-stripped to just "Version" in the config hierarchy, colliding with PlatformConfig.Version.

Issue #1338, opened 2026-06-13
[Tools] WebSearchTool count argument throws on non-integer or out-of-range JSON numbers


sytone:
## Summary
`WebSearchTool.ReadInt` parses the optional `count` argument with `el.GetInt32()` for any JSON number:
```csharp
JsonElement { ValueKind: JsonValueKind.Number } el => el.GetInt32(),
```
If the model passes a non-integer JSON number (e.g. `"count": 3.5`) or a value outside `Int32` range (e.g. `"count": 99999999999`), `GetInt32()` throws `FormatException` / `InvalidOperationException`. This exception is thrown from `PrepareArgumentsAsync` — **before** `ExecuteAsync`'s try/catch — so it escapes as an unhandled tool-preparation error instead of being handled gracefully like every other invalid input in this tool.

## Impact
- A malformed `count` (easy for a model to emit) turns a recoverable "please use a valid count" situation into a hard tool failure / surfaced exception.
- Inconsistent with the rest of `WebSearchTool`, which returns friendly `TextResult(...)` errors for bad input rather than throwing.
- `priority:medium` robustness bug — not a security issue, but a correctness/UX gap.

## Reproduction
Invoke `web_search` with:
```json
{ "query": "test", "count": 3.5 }
```
`PrepareArgumentsAsync` throws `FormatException` from `GetInt32()`.

## Proposed fix
Make `ReadInt` tolerant of non-integer / out-of-range JSON numbers instead of throwing:
- For `JsonValueKind.Number`, use `TryGetInt32` (or read as double and truncate/clamp) and return `null` when it is not a valid int32, so the existing `count` clamping logic (`Math.Min(requestedCount, _config.MaxResults)` with the `requestedCount > 0` guard) simply falls back to the configured default.
- Keep behavior identical for valid integers.

## Acceptance criteria
- `count` values that are not valid int32 (fractional, out-of-range, non-numeric strings) no longer throw; the tool falls back to the configured `MaxResults` default.
- Valid integer `count` continues to clamp to `[1, MaxResults]` as today.
- `query` validation is unchanged.
- Unit tests cover: integer count (clamped), fractional count (fallback, no throw), out-of-range count (fallback, no throw), non-numeric string count (fallback), omitted count (default).

## Target files
- `src/extensions/BotNexus.Extensions.WebTools/WebSearchTool.cs`
- `tests/extensions/BotNexus.Extensions.WebTools.Tests/` (new test coverage)

[...]

PR #1412, opened 2026-06-14
fix(portal): tolerate non-integer delay seconds in tool-call description renderer

### LINKED ISSUE #1411: [Portal] ToolDescriptionFormatter.FormatDelay throws on non-integer or out-of-range seconds
`ToolDescriptionFormatter.FormatDelay` (the portal chat-panel tool-call renderer) calls `JsonElement.GetInt32()` directly on the `seconds` argument, and the throw it can produce is **not** caught by the formatter's guard.

## Where

`src/extensions/BotNexus.Extensions.Channels.SignalR.BlazorClient/Components/ToolDescriptionFormatter.cs`

```csharp
private static string FormatDelay(JsonElement? args)
{
    if (args is null) return "Delay";
    if (args.Value.TryGetProperty("seconds", out var sec) && sec.ValueKind == JsonValueKind.Number)
        return $"Delay {sec.GetInt32()}s";   // L183 — throws on non-int / out-of-range
    return "Delay";
}
```

## Root cause

`FormatDescription` (L20) wraps only `JsonDocument.Parse(...)` in `try { ... } catch (JsonException)` (L24–31). The `switch` that dispatches to `FormatDelay` runs **outside** that `try`, and `FormatDelay` itself has no guard.

`JsonElement.GetInt32()` does **not** throw `JsonException`. It throws:
- `FormatException` when the JSON number is fractional (e.g. `{"seconds": 1.5}`)
- `InvalidOperationException` / `OverflowException` when the value is out of `Int32` range (e.g. `{"seconds": 9999999999}`)

The `ValueKind == JsonValueKind.Number` check passes for all of these (a fractional or oversized number is still `JsonValueKind.Number`), so the guard does not prevent the call. None of those exceptions are caught → they propagate out of the chat-panel render.

## Impact

Same `GetInt32()`-before-validation bug class as #1338 / #1362 / #1363 / #1365, but on the **portal display** path rather than a tool-execution path:

- A `delay` tool call with a fractional or out-of-`Int32`-range `seconds` argument crashes the tool-description renderer in the SignalR Blazor chat panel.
- `DelayTool` clamps the actual executed value, so this is **display-only** — it does not affect execution. Severity is low: it requires a malformed/adversarial `delay` argument and only breaks the in-flight render of that one tool call's label.

## Fix

Use the saturating/tolerant pattern already adopted in the fixed tools (`TryGetInt32` → fall back), e.g.:

```csharp
if (args.Value.TryGetProperty("seconds", out var sec) && sec.ValueKind == JsonValueKind.Number)
    return sec.TryGetInt32(out var s) ? $"Delay {s}s" : "Delay";
```

(or widen the existing guard to also catch the non-`JsonException` types around the dispatch, but the `TryGetInt32` fix is the minimal, targeted one.)

## Acceptance criteria

- `FormatDelay` renders a stable label for fractional and out-of-`Int32`-range `seconds` values instead of throwing.
- A unit test in the BlazorClient test project covers: integer `seconds` (normal), fractional `seconds`, out-of-range `seconds`, and missing/non-number `seconds` → never throws.
- No behavior change for valid integer inputs.

sytone:
Closes #1411

## Changes
`ToolDescriptionFormatter.FormatDelay` (the SignalR Blazor chat-panel tool-call renderer) called `JsonElement.GetInt32()` directly on the `seconds` argument. `GetInt32()` throws `FormatException` on a fractional JSON number and `InvalidOperationException`/`OverflowException` on an out-of-`Int32`-range number — none of which are `JsonException`, so the formatter's parse guard (`try { JsonDocument.Parse } catch (JsonException)`) does **not** catch them. A `delay` tool call with `{"seconds": 1.5}` or `{"seconds": 9999999999}` would crash the tool-description render.

Fix: use `TryGetInt32` and fall back to the plain `"Delay"` label when the value is not a valid `Int32` — the same tolerant pattern already adopted for the `GetInt32()`-before-validation class (#1338, #1362, #1363).

This is display-only — `DelayTool` already clamps the executed value — so it is a low-severity crash on the in-flight render of a malformed `delay` argument.

## Tests
Added 3 cases to `ToolDescriptionFormatterTests` (the existing in-slnx project):
- `Delay_WithFractionalSeconds_FallsBackWithoutThrowing`
- `Delay_WithOutOfRangeSeconds_FallsBackWithoutThrowing`
- `Delay_WithNonNumberSeconds_FallsBack`

The first two fail against the old code (they throw). `BlazorClient.Tests` 540/540, Architecture 178, Scenarios 29, Gateway 2868 — all green. Full solution build clean (0 warnings, 0 errors).

## Merge Notes
Touches only `ToolDescriptionFormatter.cs` (+ its test file) in `BotNexus.Extensions.Channels.SignalR.BlazorClient`. No file overlap with any other open PR — safe to merge in any order.

### DIFF
--- a/src/extensions/BotNexus.Extensions.Channels.SignalR.BlazorClient/Components/ToolDescriptionFormatter.cs
+++ b/src/extensions/BotNexus.Extensions.Channels.SignalR.BlazorClient/Components/ToolDescriptionFormatter.cs
@@ -179,8 +179,11 @@ public static class ToolDescriptionFormatter
     private static string FormatDelay(JsonElement? args)
     {
         if (args is null) return "Delay";
-        if (args.Value.TryGetProperty("seconds", out var sec) && sec.ValueKind == JsonValueKind.Number)
-            return $"Delay {sec.GetInt32()}s";
+        if (args.Value.TryGetProperty("seconds", out var sec) && sec.ValueKind == JsonValueKind.Number
+            && sec.TryGetInt32(out var seconds))
+        {
+            return $"Delay {seconds}s";
+        }
         return "Delay";
     }
 
--- a/tests/extensions/BotNexus.Extensions.Channels.SignalR.BlazorClient.Tests/ToolDescriptionFormatterTests.cs
+++ b/tests/extensions/BotNexus.Extensions.Channels.SignalR.BlazorClient.Tests/ToolDescriptionFormatterTests.cs
@@ -215,6 +215,30 @@ public sealed class ToolDescriptionFormatterTests
     }
 
     [Fact]
+    public void Delay_WithFractionalSeconds_FallsBackWithoutThrowing()
+    {
+        // GetInt32() throws FormatException on a fractional JSON number; that is not a
+        // JsonException, so the formatter's parse guard would not catch it. It must degrade.
+        var result = ToolDescriptionFormatter.FormatDescription("delay", """{"seconds": 1.5}""");
+        Assert.Equal("⏳ Delay", result);
+    }
+
+    [Fact]
+    public void Delay_WithOutOfRangeSeconds_FallsBackWithoutThrowing()
+    {
+        // 9999999999 is a valid JSON number but out of Int32 range; GetInt32() would throw.
+        var result = ToolDescriptionFormatter.FormatDescription("delay", """{"seconds": 9999999999}""");
+        Assert.Equal("⏳ Delay", result);
+    }
+
+    [Fact]
+    public void Delay_WithNonNumberSeconds_FallsBack()
+    {
+        var result = ToolDescriptionFormatter.FormatDescription("delay", """{"seconds": "soon"}""");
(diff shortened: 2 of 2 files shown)


sytone:
## CI Health Check -- PR #1412

| Check | Status |
|-------|--------|
| impacted-tests | pending |
| CodeQL | pending |

**Branch:** `fix/tooldesc-delay-format` | **Behind main:** 0 commits | **Mergeable:** MERGEABLE

**Actions taken:**
- Merged main into branch and pushed (was behindBy 13-14 after the merge wave); CI re-triggered on the merge commit.

**Blockers for Jon:**
- None

---
- 2026-06-14 06:09 UTC
- 2026-06-14 02:06 UTC
---
*Farnsworth (automated CI monitor) -- [BotNexus](https://github.com/Sytone/botnexus) -- Last updated: 2026-06-14 06:09 UTC*

PR #1434, opened 2026-06-14
fix(docker): install curl so container healthcheck reports healthy

### LINKED ISSUE #1432: Dockerfile HEALTHCHECK uses curl which is absent from the aspnet:10.0 base image (container always reports unhealthy)
## Summary

The `Dockerfile` `HEALTHCHECK` uses `curl`, which is **not present** in the `mcr.microsoft.com/dotnet/aspnet:10.0` runtime base image. As a result the health probe exits `1` on every interval and the container is permanently reported as **`unhealthy`**, even though the gateway is fully up and `GET /health` returns `200 {"status":"ok"}`.

Found during daily build validation (2026-06-14, main `e5ea512e`) — the **first clean end-to-end container boot of `main`** (the #1360 boot crash is now fixed). The gateway itself is healthy; only the Docker `HEALTHCHECK` mechanism is broken.

## Repro

```bash
docker build -t botnexus:test .
docker run -d --name bn -p 5199:5000 -v <config>:/app/config botnexus:test
# Gateway works:
curl http://localhost:5199/health        # -> 200 {"status":"ok",...}
# But Docker health is broken:
docker inspect -f '{{.State.Health.Status}}' bn
# -> unhealthy
docker inspect -f '{{range .State.Health.Log}}{{.ExitCode}}: {{.Output}}{{end}}' bn
# -> 1: /bin/sh: 1: curl: not found
docker exec bn sh -c 'which curl || echo NO_CURL'   # -> NO_CURL
docker exec bn sh -c 'which wget || echo NO_WGET'   # -> NO_WGET
```

The current healthcheck line:
```dockerfile
HEALTHCHECK --interval=10s --timeout=5s --start-period=30s --retries=5 \
    CMD curl -f http://localhost:5000/health || exit 1
```

## Impact

- **`docker run`/compose/orchestrators that gate on container health** (k8s readiness via the docker health status, swarm, `--health`, `depends_on: condition: service_healthy`) will treat the container as never-ready → may refuse traffic, fail deploys, or restart-loop a perfectly working gateway.
- Confined to the Docker `HEALTHCHECK` mechanism — the HTTP `/health` endpoint and the gateway are unaffected.
- The aspnet:10.0 base ships **neither `curl` nor `wget`**; only `dotnet` is available.

## Suggested fix (any one)

1. **Install curl in the runtime stage** (smallest diff, explicit dependency):
   ```dockerfile
   RUN apt-get update && apt-get install -y --no-install-recommends curl \
       && rm -rf /var/lib/apt/lists/*
   ```
2. **Use a dotnet-based probe** — a tiny `HEALTHCHECK CMD` that shells a managed HTTP check (no extra package), or ship a minimal health-probe tool.
3. **Switch to `wget`** after installing it (same apt cost as curl).

Option 1 is the most common and lowest-risk. Whichever is chosen, add a build-validation assertion that `docker inspect ... .State.Health.Status` reaches `healthy` so this can't regress silently.

## Labels
`bug`, `integration-test`

sytone:
Closes #1432

## Changes
- **`Dockerfile`**: install `curl` in the runtime stage so the `HEALTHCHECK` probe works. The `mcr.microsoft.com/dotnet/aspnet:10.0` base image ships neither `curl` nor `wget` (only `dotnet`), so the existing `HEALTHCHECK curl -f http://localhost:5000/health` exited `1` on every interval and the container was reported `unhealthy` forever even though `GET /health` returns `200`.
- **`tests/architecture/BotNexus.Architecture.Tests/DockerHealthcheckArchitectureTests.cs`** (new): Docker-free static fitness function. If the `Dockerfile` (or `docker-compose.yml`) `HEALTHCHECK` depends on a `curl`/`wget` binary, the `Dockerfile` must install it. Includes vacuity + positive pins (the broken pre-fix shape is detected; the fixed shape is accepted). This is the regression guard the issue asked for, runnable in CI where Docker is not available.
- **`docs/development/container-integration-testing.md`**: new "Container Health Probe" section documenting the curl requirement, the `docker inspect ... .State.Health.Status` check, and the architecture guard.

## Why install curl (vs other options)
The issue listed three fixes (install curl / dotnet-native probe / install wget). Option 1 (install curl) is the lowest-risk, most-common approach and matches the existing `HEALTHCHECK` and `docker-compose.yml` `test` line, so no further drift. `BotNexus.Probe` is an OTLP/session-diagnostics service, not a health-check CLI, so a dotnet-native probe would mean shipping a new binary — overkill for a one-line healthcheck. The architecture fence is binary-agnostic, so a future dotnet-native switch is still permitted.

## docker-compose.yml
No change needed: compose builds the same Dockerfile image, which now ships `curl`, so its `["CMD","curl",...]` healthcheck works. The new architecture test also guards the compose healthcheck against drift.

## Verification
- Full solution build clean (0 warnings, 0 errors).
- `test-impacted.ps1`: Architecture **185** (+5 new), Scenarios **29** — all pass.
- The fence FAILS against the original (curl, no install) shape and PASSES on the fixed shape (proven by the in-test synthetic vacuity + positive pins).

## Merge Notes
Standalone — only touches `Dockerfile`, a new Architecture.Tests file, and a docs page. Zero overlap with any open PR. Safe to merge in any order.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,6 +15,14 @@ RUN dotnet publish src/gateway/BotNexus.Gateway.Api/BotNexus.Gateway.Api.csproj 
 FROM mcr.microsoft.com/dotnet/aspnet:10.0 AS runtime
 WORKDIR /app
 
+# The aspnet:10.0 base image ships neither curl nor wget (only dotnet), so the
+# HEALTHCHECK below needs curl installed explicitly. Without it the probe exits 1
+# on every interval and the container is reported unhealthy forever even though
+# GET /health returns 200. See issue #1432.
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends curl \
+    && rm -rf /var/lib/apt/lists/*
+
 COPY --from=build /app/publish .
 
 # BOTNEXUS_HOME is the config directory; mount a volume here with your config.json (and optionally auth.json).
--- a/docs/development/container-integration-testing.md
+++ b/docs/development/container-integration-testing.md
@@ -74,6 +74,34 @@ docker logs -f botnexus-gateway
 docker stop botnexus-gateway && docker rm botnexus-gateway
 ```
 
+## Container Health Probe
+
+The `Dockerfile` declares a Docker `HEALTHCHECK` that probes `GET /health` with `curl`.
+The `mcr.microsoft.com/dotnet/aspnet` runtime base image ships **neither `curl` nor `wget`**
+(only `dotnet`), so the runtime stage explicitly installs `curl`:
+
+```dockerfile
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends curl \
+    && rm -rf /var/lib/apt/lists/*
+```
+
+Without this step the probe exits `1` on every interval and the container is reported
+`unhealthy` forever — even though the gateway is fully up and `GET /health` returns `200`.
+Orchestrators that gate on container health (`depends_on: condition: service_healthy`,
+swarm, k8s readiness via the docker health status) then treat a working gateway as never-ready.
+
+Verify the container reaches `healthy` after boot:
+
+```bash
+docker inspect -f '{{.State.Health.Status}}' botnexus-gateway   # -> healthy
+```
+
+This is guarded by `DockerHealthcheckArchitectureTests` (a Docker-free static fitness
+function): if the `Dockerfile` or `docker-compose.yml` HEALTHCHECK depends on `curl`/`wget`,
+the `Dockerfile` must install that binary. A future switch to a dotnet-native probe (no
+external binary) would satisfy the guard without the apt layer.
+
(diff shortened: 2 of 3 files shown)


sytone:
## CI Health Check -- PR #1434

| Check | Status |
|-------|--------|
| impacted-tests | pending |
| CodeQL | pending |

**Branch:** `fix/docker-healthcheck-curl` | **Behind main:** 0 commits | **Mergeable:** MERGEABLE

**Actions taken:**
- Synced branch with main after merge wave (was behindBy 8); CI re-triggered.

**Blockers for Jon:**
- None

---
- 2026-06-14 19:01 UTC
---
*Farnsworth (automated CI monitor) -- [BotNexus](https://github.com/Sytone/botnexus) -- Last updated: 2026-06-14 19:01 UTC*

PR #1820, opened 2026-07-08
feat(gateway): add conversation-level model, thinking, and context override

sytone:
Closes #1706

## Changes
Implements PBI5 of epic #1701 — the **conversation-override layer** (top precedence) of the 3-layer model/thinking/context override stack. The merged `ModelOverrideResolver` (PBI3 #1810) already consumes this layer; this PR populates and persists it.

- `Conversation` domain model gains `ModelOverride` (string?), `ThinkingOverride` (string?), `ContextWindowOverride` (int?) — mirroring the existing `Instructions`/`TodoJson` conversation-scoped pattern.
- `SqliteConversationStore` + `ConversationRowMapper`: three new nullable columns (`model_override`, `thinking_override`, `context_window_override`), additive migration, full round-trip.
- `ConversationsController` + DTOs: API to set/clear the overrides; `ConversationOverrideValidator` (in Providers.Core, the resolver's home) validates values against the selected model's capability set.
- `InProcessIsolationStrategy` feeds the conversation layer into the resolver as **top precedence**, reusing the existing memoised conversation-id resolution (#1382 Finding 2) so no second DB round-trip is added.
- `BuiltInCommandContributor`: `/model`, `/reasoning`, `/context` slash-commands. Portal per-conversation picker (`ConversationOverridePanel`).
- Docs: `docs/user-guide/conversations.md`.

## Tests
TDD, happy + sad paths: `ConversationModelOverrideTests` (override beats agent, persists across store round-trip, clears back to agent), `ConversationOverrideValidatorTests` (invalid-for-model rejected), `ConversationsControllerOverrideTests`, `ConversationOverridePanelTests`, plus updated `ConversationRowMapperTests` fixtures for the new columns. `test-impacted.ps1` green (Gateway 3521, Conversations 280, BlazorClient 675, Providers.Core 391, Architecture fence 5/5).

## Merge Notes
Disjoint from #1705 (agent-level) except both touch `InProcessIsolationStrategy.CreateAsync` (the resolver call-site) — a trivial textual merge (agent layer vs conversation layer arguments to the same `Resolve` call). Conversation override is top precedence over the agent default. PBI6 #1707 (dynamic-model capabilities) builds on both.

PR #1855, opened 2026-07-09
fix(gateway): keep agent loop alive across pre-compaction memory flush

sytone:
Closes #1845

## Changes

Pre-compaction memory flush no longer abandons an agent's in-flight task.

**Root cause:** the flush prompt was injected as a normal steered `AgentCoreUserMessage`. When compaction fired mid-run, the model answered the flush turn with plain text (write memory + status) → `hasMoreToolCalls=false` and the steering queue empty → the loop's `while (hasMoreToolCalls || pendingMessages.Count > 0)` condition went false → the loop **exited**, abandoning the original in-flight work.

**Fix:** thread a `DeferWhileBusy` flag through the flush path so a system-injected side turn is held aside while the run still has pending tool calls, then released only at a genuine idle boundary.

- `UserMessage.DeferWhileBusy` (new `init` flag, default `false` → unchanged normal-steer behaviour)
- `IAgentHandle.SteerDeferrableAsync` (default-implemented to fall back to `SteerAsync`)
- `InProcessAgentHandle.SteerDeferrableAsync` marks the steered message `DeferWhileBusy = true`
- `PreCompactionMemoryFlusher` now calls `SteerDeferrableAsync` instead of `SteerAsync`
- `AgentLoopRunner`: mid-loop drain extracts any `DeferWhileBusy` messages into a `deferredMessages` holdout (`ExtractDeferWhileBusy`); the outer `while` keeps running while deferred work remains; deferred messages are released into `pendingMessages` only once `!hasMoreToolCalls && pendingMessages.Count == 0`. The **initial** steering poll is left untouched, so a flush queued while the agent is genuinely idle behaves exactly as before.

## Tests

New `DeferWhileBusySteeringTests`:
- `DeferWhileBusy_InjectedMidFlight_DoesNotConsumeLoopContinuation` — asserts a mid-flight flush is NOT injected while a tool call is pending, and IS released once the run reaches idle (loop continues the original task past the flush).
- `DeferWhileBusy_WhenAgentIdleAtInject_BehavesLikeNormalSteer` — no behaviour change when idle.

Full `test-impacted.ps1` green: Agent.Core.Tests 237 (incl. 28 existing SteeringLoopTests + 2 new), Architecture, Scenarios, and all impacted suites pass.

## Merge Notes

File footprint (AgentLoopRunner / AgentMessage / IAgentHandle / InProcessIsolationStrategy / PreCompactionMemoryFlusher) is disjoint from the other open PRs (#1820 conversation-override, #1843 mobile-hub-reset, #1846 skill-telemetry, #1847 pre-commit hook). Safe to merge in any order.


sytone:
## CI Health Check -- PR #1855

| Check | Status |
|-------|--------|
| impacted-tests | pending |
| CodeQL | pending |

**Branch:** `fix/compaction-flush-loop-continuation` | **Behind main:** 0 commits | **Mergeable:** MERGEABLE

**Actions taken:**
- fixed impacted-tests: updated PreCompactionMemoryFlusherTests to verify SteerDeferrableAsync (the PR's new call) instead of the old SteerAsync; all 10 flusher tests pass locally

**Blockers for Jon:**
- None

---
- 2026-07-09 07:04 UTC
---
*Farnsworth (automated CI monitor) -- [BotNexus](https://github.com/Sytone/botnexus) -- Last updated: 2026-07-09 07:04 UTC*

